PR #18, opened 2015-03-22
CRLF support

### LINKED ISSUE #14: CRLF support
Thanks for your quick work on issue #11. I've continued integrating Cmark with my project.

I've found that when parsing files that use CRLF line endings with Cmark, block quotes extend into subsequent paragraphs, text after lists stays indented, and there are problems with fenced code.

I think the biggest problem is that the test cases don't seem to exercise alternate line endings at all. I'd suggest that _all_ of the test cases should be automatically converted to each type of line ending (LF, CR, and CRLF) during `make test` and Cmark should be expected to pass all of them.

Perhaps there should also be some test cases for a single file brokenly mixing different line endings. This should be possible to handle reasonably intelligently.

I got CRLF support mostly working with some quick patches to blocks.c, which I can post if it would help. I didn't try digging into the tests.

btrask:
Fixes #14.

Note that this has some failing tests due to #17. However I don't think any new problems were introduced with these changes.

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -213,10 +213,10 @@ finalize(cmark_parser *parser, cmark_node* b)
 	           (b->type == NODE_HEADER && b->as.header.setext)) {
 		b->end_line = parser->line_number;
 		b->end_column = parser->curline->size;
-		if (b->end_column && parser->curline->ptr[b->end_column-1] == '\n')
-			b->end_column -= 1;
-		if (b->end_column && parser->curline->ptr[b->end_column-1] == '\r')
-			b->end_column -= 1;
+		if (b->end_column && parser->curline->ptr[b->end_column - 1] == '\n')
+			b->end_column--;
+		if (b->end_column && parser->curline->ptr[b->end_column - 1] == '\r')
+			b->end_column--;
 	} else {
 		b->end_line = parser->line_number - 1;
 		b->end_column = parser->last_line_length;
@@ -895,11 +895,11 @@ S_process_line(cmark_parser *parser, const unsigned char *buffer, size_t bytes)
 finished:
 	parser->last_line_length = parser->curline->size;
 	if (parser->last_line_length &&
-	    parser->curline->ptr[parser->last_line_length-1] == '\n')
-		parser->last_line_length -= 1;
+	    parser->curline->ptr[parser->last_line_length - 1] == '\n')
+		parser->last_line_length--;
 	if (parser->last_line_length &&
-	    parser->curline->ptr[parser->last_line_length-1] == '\r')
-		parser->last_line_length -= 1;
+	    parser->curline->ptr[parser->last_line_length - 1] == '\r')
+		parser->last_line_length--;
 
 	cmark_strbuf_clear(parser->curline);
 
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -145,7 +145,7 @@ static void remove_trailing_blank_lines(cmark_strbuf *ln)
 	}
 
 
-	for(i = 0; i < ln->size; ++i) {
+	for(; i < ln->size; ++i) {
 		c = ln->ptr[i];
 
 		if (c != '\r' && c != '\n')
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -90,9 +90,6 @@ static bool is_blank(cmark_strbuf *s, int offset)
 	while (offset < s->size) {
 		switch (s->ptr[offset]) {
 		case '\r':
-			if (s->ptr[offset + 1] == '\n')
-				offset++;
-			return true;
 		case '\n':
 			return true;
 		case ' ':
@@ -150,10 +147,6 @@ static void remove_trailing_blank_lines(cmark_strbuf *ln)
 
 		if (c != '\r' && c != '\n')
 			continue;
-
-		// Don't cut a CRLF in half
-		if (c == '\r' && i+1 < ln->size && ln->ptr[i+1] == '\n')
-			++i;
 
 		cmark_strbuf_truncate(ln, i);
 		break;
@@ -568,10 +561,13 @@ S_process_line(cmark_parser *parser, const unsigned char *buffer, size_t bytes)
 
 	// Add a newline to the end if not present:
 	// TODO this breaks abstraction:
-	// Note: we assume output is LF-only
-	if (parser->curline->ptr[parser->curline->size - 1] != '\n') {
-		cmark_strbuf_putc(parser->curline, '\n');
+	if (parser->curline->ptr[parser->curline->size - 1] == '\n') {
+		cmark_strbuf_truncate(parser->curline, parser->curline->size - 1);
 	}
+	if (parser->curline->ptr[parser->curline->size - 1] == '\r') {
+		cmark_strbuf_truncate(parser->curline, parser->curline->size - 1);
+	}
+	cmark_strbuf_putc(parser->curline, '\n');
 	input.data = parser->curline->ptr;
 	input.len = parser->curline->size;
 
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -89,6 +89,10 @@ static bool is_blank(cmark_strbuf *s, int offset)
 {
 	while (offset < s->size) {
 		switch (s->ptr[offset]) {
+		case '\r':
+			if (s->ptr[offset + 1] == '\n')
+				offset++;
+			return true;
 		case '\n':
 			return true;
 		case ' ':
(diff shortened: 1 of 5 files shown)


src/blocks.c
```diff
@@ -204,9 +219,11 @@ finalize(cmark_parser *parser, cmark_node* b)
...
 	           (b->type == NODE_HEADER && b->as.header.setext)) {
 		b->end_line = parser->line_number;
-		b->end_column = parser->curline->size -
-		                (parser->curline->ptr[parser->curline->size - 1] == '\n' ?
-		                 1 : 0);
+		b->end_column = parser->curline->size;
+		if (b->end_column && parser->curline->ptr[b->end_column-1] == '\n')
+			b->end_column -= 1;
+		if (b->end_column && parser->curline->ptr[b->end_column-1] == '\r')
+			b->end_column -= 1;
```
nwellnhof: Shouldn't the second `if` statement only be executed if the first condition is true?

src/blocks.c
```diff
@@ -232,19 +249,28 @@ finalize(cmark_parser *parser, cmark_node* b)
 		} else {
 
 			// first line of contents becomes info
-			firstlinelen = cmark_strbuf_strchr(&b->string_content, '\n', 0);
+			for (pos = 0; pos < b->string_content.size; ++pos) {
+				if (b->string_content.ptr[pos] == '\r' ||
+				    b->string_content.ptr[pos] == '\n')
+					break;
+			}
```
nwellnhof: This seems to break on every `\r` even if not followd by `\n`. Is this the intention?

src/blocks.c
```diff
@@ -467,13 +493,22 @@ S_parser_feed(cmark_parser *parser, const unsigned char *buffer, size_t len,
...
+			eol = NULL;
+
 		if (eol) {
-			line_len = eol + 1 - buffer;
+			if (eol < end && *eol == '\r')
+				eol++;
+			if (eol < end && *eol == '\n')
+				eol++;
+			line_len = eol - buffer;
 		} else if (eof) {
```
nwellnhof: This seems to break on every `\r` even if not followed by `\n`. Is this the intention?

src/blocks.c
```diff
@@ -467,13 +493,22 @@ S_parser_feed(cmark_parser *parser, const unsigned char *buffer, size_t len,
 	const unsigned char *end = buffer + len;
 
 	while (buffer < end) {
```
nwellnhof: This doesn't find `\r\n` if it lies across a buffer boundary.

btrask:
The checks are carefully set up to handle CR, LF, and CRLF at the same time. Basically you check for CR first, then increment, then check for LF. So if it's CRLF you get both. In some cases it's scanning backwards so it checks LF first.

CR-only line endings are used on some systems (not as common any more, [according to Wikipedia](https://en.wikipedia.org/wiki/Newline#Representations)). I didn't see any harm in supporting them. (LFCR is unsupported since I've never seen it used.)

I think you're right about it not handling buffer boundaries though.

nwellnhof:
OK, so the intention is to support single CRs as well. I don't have a problem with that.

@jgm I think the spec could also be improved in that regard. It currently says:

> A line ending is, depending on the platform, a newline (U+000A), carriage return (U+000D), or carriage return + newline.

I don't like the "depending on the platform" part. This basically means that CommonMark is interpreted differently depending on the platform. Essentially, this makes it impossible to reliably transfer files between platforms. Also, what is the platform in the context of a web application? Can't we just drop that part and have a single definition for newlines across all platforms?

jgm:
+++ Nick Wellnhofer [Mar 24 15 08:15 ]:
>   [1]@jgm I think the spec could also be improved in that regard. It
>   currently says:
>
>     A line ending is, depending on the platform, a newline (U+000A),
>     carriage return (U+000D), or carriage return + newline.
>
>   I don't like the "depending on the platform" part. This basically means
>   that CommonMark is interpreted differently depending on the platform.
>   Essentially, this makes it impossible to reliably transfer files
>   between platforms. Also, what is the platform in the context of a web
>   application? Can't we just drop that part and have a single definition
>   for newlines across all platforms?

I'm sympathetic with this.  To explain what my thinking was:  the parser operates on a sequence of characters separated into lines.  How these characters and line breaks are encoded in bytes is a concrete implementation detail that the parser does not specify.  So, you might have latin1 encoding and LFs as line breaks, or UTF8 encoding and CRLF line breaks.

But there are really three levels of abstraction we could aim for:

1.  Define the spec over a sequence of bytes, specifying everything, including what counts as a line break and how characters are encoded.

2.  Define the spec over a sequence of unicode code points, specifying what counts as a line break in terms of these, but not specifying how characters are encoded as bytes.

3.  Define the spec over a sequence of *lines*, each of which is a sequence of unicode code points, not specifying how characters are encoded as bytes *or* how a stream of characters is divided into lines.

I think what I had in mind was roughly 3, but the spec didn't make this too clear.  We could switch to 2 or even 1. This may need further discussion on talk.commonmark.org.

Further question:  if we said explicitly that any of these combinations counted as a line ending, would we allow these styles to be mixed in the same document?

nwellnhof:
@jgm My proposal is to aim for the second approach.

I think the first approach doesn't make much sense. It limits the CommonMark format to a single encoding (UTF-8).

And yes, I would make any of CR (not followed by LF), LF (not preceded by CR) and CR+LF count as line ending, also allowing different line endings in the same document. This seems like the only workable approach to cross-platform line endings. It should match the behavior this pull request implements. Not sure about the JS implementation, though.

btrask:
I found some bugs with underlined headers. I think what's happening is that `\r`s and `\n`s are getting counted as two separate line breaks, and then getting merged. Fixing this might take some more effort because cmark generally assumes that line breaks are a single byte.

I looked into getting the tests running with different line endings but I haven't figured it out yet. Without tests it's hard to tell how much is broken.

jgm:
Tests are essential.  For something quick and dirty,
can't you just do a regex replace on the markdown content
just before it's passed to the program (line 39 of spec_test.py)?

btrask:
I'll look into it, thanks.

btrask:
Fixed three bugs, now all the tests pass!

The tests themselves are just hacked up, right now they are CRLF only. I also tested with CR and that works too.

I'm not sure why the Travis build is failing...

btrask:
Fixed :)

btrask:
@jgm have you had a chance to look at this yet?

I can try to work on integrating the tests better if that's what you're waiting for.

jgm:
+++ Ben Trask [Apr 21 15 14:59 ]:
>@jgm have you had a chance to look at this yet?

No, I've just been busy.  I will get to it, though.

jgm:
@btrask I've had a look and added some line comments.  Aside from those things, it looks good.  Thanks for doing this.

@nwellnhof  I also checked commonmark.js, and it seems to behave the same way as cmark does with this change.  So, we just need to revise the spec.  Can you add an issue on that to jgm/CommonMark?

btrask:
Thanks for getting back to this!

Regarding the calls to `cmark_strbuf_truncate`, it's a plain function call and the data is already going to be cached. I think the overhead is probably trivial and writing it more straightforwardly reduces the risk of bugs. But I'll try to come up with an alternate version that you might like better.

Regarding the `-= 1` style, I wasn't sure. I'll change it back.

You're right that the tests currently *only* check CRLF input. That is a hack that I mentioned above but I should've put a comment in the code too. I honestly wasn't sure the best way to restructure the tests to check all three modes.

jgm:
+++ Ben Trask [May 26 15 14:55 ]:
>Thanks for getting back to this!
>
>Regarding the calls to `cmark_strbuf_truncate`, it's a plain function call and the data is already going to be cached. I think the overhead is probably trivial and writing it more straightforwardly reduces the risk of bugs. But I'll try to come up with an alternate version that you might like better.

We're really going for maximum performance, so avoiding even one
unnecessary function call that gets called every line might help.

>Regarding the `-= 1` style, I wasn't sure. I'll change it back.

It's not that important. I only commented because your reason for
changing it was to match existing style.

>You're right that the tests currently *only* check CRLF input. That is a hack that I mentioned above but I should've put a comment in the code too. I honestly wasn't sure the best way to restructure the tests to check all three modes.

You can let me take care of that.  I was thinking we'd run all
the tests with all three line ending patterns.  But now I'm thinking
that will take too long, and that it isn't necessary.  Instead, I
may add a separate test for line ending handling.

btrask:
Here is an alternate version.

```c
int trim = 0;
bool cr = false;
bool lf = false;

[...]

// Add a newline to the end if not present:
// TODO this breaks abstraction:
if (parser->curline->size > trim &&
    parser->curline->ptr[parser->curline->size - 1 - trim] == '\n') {
	trim += 1;
	lf = true;
}
if (parser->curline->size > trim &&
    parser->curline->ptr[parser->curline->size - 1 - trim] == '\r') {
	trim += 1;
	cr = true;
}
if (cr) {
	cmark_strbuf_truncate(parser->curline, parser->curline->size - trim);
}
if (cr || !lf) {
	cmark_strbuf_putc(parser->curline, '\n');
}
```

For the average case with LF input, it saves two function calls per line. For CRLF input, it saves one function call per line. I haven't done any benchmarks but I strongly suspect any performance difference will be drowned by noise (maybe one microsecond for a large document).

I'll leave the tests to you. Thanks.

jgm:
Note to self:  I have merged this and pushed a branch jgm/line-endings.
I still need to look carefully at the changes and write proper tests.

btrask:
Great work, thank you!

PR #21, opened 2015-03-27
Update list of buggy re2c versions

nwellnhof:
re2c 0.14.2 fixed the issue we were hitting.


jgm:
Great, thanks for intervening with them.

PR #24, opened 2015-03-31
Install compiler-provided system runtime libraries.

yangcha:
On windows, cmark.exe and cmark.dll depend on msvcr***.dll, this line will automatically install the system runtime libraries.


jgm:
Thanks!

PR #34, opened 2015-05-06
Make it possible to use NULL in setters.

elibarzilay:
This makes it possible to use null in `cmark_node_set_literal()`,
`cmark_node_set_fence_info()`, `cmark_node_set_url()` and
`cmark_node_set_title()`.  Note that to make it work in the first two I
changed `cmark_chunk_set_cstr()` --- it's used in one other place (in
`commonmark.c`) which seems like it never sends it a NULL.

### DIFF
--- a/src/chunk.h
+++ b/src/chunk.h
@@ -78,10 +78,16 @@ static inline void cmark_chunk_set_cstr(cmark_chunk *c, const char *str)
 	if (c->alloc) {
 		free(c->data);
 	}
-	c->len   = strlen(str);
-	c->data  = (unsigned char *)malloc(c->len + 1);
-	c->alloc = 1;
-	memcpy(c->data, str, c->len + 1);
+	if (str == NULL) {
+		c->len   = 0;
+		c->data  = NULL;
+		c->alloc = 0;
+	} else {
+		c->len   = strlen(str);
+		c->data  = (unsigned char *)malloc(c->len + 1);
+		c->alloc = 1;
+		memcpy(c->data, str, c->len + 1);
+	}
 }
 
 static inline cmark_chunk cmark_chunk_literal(const char *data)
--- a/src/node.c
+++ b/src/node.c
@@ -560,7 +560,8 @@ cmark_node_set_url(cmark_node *node, const char *url)
 	case NODE_LINK:
 	case NODE_IMAGE:
 		free(node->as.link.url);
-		node->as.link.url = (unsigned char *)S_strdup(url);
+		node->as.link.url =
+		    (url == NULL) ? NULL : (unsigned char *)S_strdup(url);
 		return 1;
 	default:
 		break;
@@ -598,7 +599,8 @@ cmark_node_set_title(cmark_node *node, const char *title)
 	case NODE_LINK:
 	case NODE_IMAGE:
 		free(node->as.link.title);
-		node->as.link.title = (unsigned char *)S_strdup(title);
+		node->as.link.title =
+		    (title == NULL) ? NULL : (unsigned char *)S_strdup(title);
 		return 1;
 	default:
 		break;


nwellnhof:
I agree that it's a good idea to not crash if NULL is passed to one of the string setters. But I'd prefer to return an error in this case and leave the string value untouched.

elibarzilay:
Actually there is a possible valid use case for this: it is currently not possible to clear out a field, you can at most set it to `""`.  For example, rendering a link with an empty title is probably different than rendering a link with no title -- so this will make it possible to get to the latter case.

elibarzilay:
Perhaps a better way to say it is that (IMO) the interface should be symmetric in that if NULL is not a valid value to set, then you shouldn't get one out from the respective getters.  IOW, it's nice if you can always send setters whatever values you get from their matching getters.

nwellnhof:
Good point. A missing title currently makes `cmark_get_title` return `NULL`. But `cmark_chunk`s used for "literal" and fence info never have a `NULL` value. If the fence info is missing, you'll simply get an empty string.

To restore symmetry, I'd prefer to make `cmark_get_title` return the empty string for missing titles.

jgm:
+++ Nick Wellnhofer [May 07 15 09:00 ]:
>   Good point. A missing title currently makes cmark_get_title return
>   NULL. But cmark_chunks used for "literal" and fence info never have a
>   NULL value. If the fence info is missing, you'll simply get an empty
>   string.
>
>   To restore symmetry, I'd prefer to make cmark_get_title return the
>   empty string for missing titles.

I'd rather not do that, as I think it might be confusing if directly accessing
the title and url, as we do in the library code, gave different results from
`cmark_get_title`.  At least, I'd worry about getting confused.

But maybe it would make sense to do that AND change the title and url fields
from `*char` to `chunk`.  That would restore uniformity and symmetry, and
we could insist on empty strings instead of NULL.

@elibarzilay points out that if we had both empty string and NULL as possible
values for title, we could distinguish between

    

and

    

But I'm not sure this is important, as these behave the same in HTML anyway.

elibarzilay:
I just got to check it now, and the current state looks like it needs some tidying: parsing a link with an empty title makes it have a NULL title, which is the same as what you get for no title.  Rendering that produces no title.  But creating a node with an empty title *does* render a `title=""`.  This makes things weird in that without my change you cannot create something that renders no title (unless you leave the field untouched, assuming it's initialized to NULL -- and that's a complication in converting these values to other languages).

So if no NULL should go in, then I'd expect them to not come out, and I'd expect empty strings to lead to the same result as missing values.

(But IMO they should be allowed, to distinguish no title from an empty one -- I'm sure that there's some renderer where the difference would makes sense.)

nwellnhof:
@elibarzilay Yes, the current behavior needs to be fixed.

You convinced me that for titles, users would expect that `cmark_node_set_title(node, NULL)` removes the title. So this should be allowed. If we allow NULL values for titles, we should support them for other properties, too. But I'd treat them more like a default value.

I think it should be OK to treat missing and empty values the same for all renderers. If we do that, I'd prefer that the getters always return the empty string instead of `NULL`.

@jgm Yes, it would be nice if the internal and external representation would always be the same, but I don't think it's a big problem. It's more important to get the external interface right.

jgm:
These are the two options, I think:

1.  Keep title and url as `unsigned char *`.  Allow them to
be set to NULL.  (Basically, your PR.)

2.  Change title and url to `chunk`.  Don't allow them to
have NULL values.  Change the HTML renderer so that a title
attribute isn't produced if the title is the empty string.

I lean slightly towards 2, on the grounds of uniformity.
(Everything else is a chunk, and can't take a NULL value,
so why these?)  If I thought there was a reason to
distinguish between the absence of a title attribute and
an empty title attribute, that might incline me towards 1.
But I can't actually think of any output formats where that
would be needed.

nwellnhof:
I prefer the second approach, too. If it's OK, I can make the necessary changes and also change the setters to allow `NULL` values.

jgm:
@nwellnhof that sounds good to me - please go ahead.

nwellnhof:
This can be closed now. Thanks again, @elibarzilay!

elibarzilay:
@nwellnhof: it looks like things are flipped now: IIUC (just by looking at the commit message), I can never get a NULL as a result from a getter, but I can use one in a setter (which will be used as an empty string)?

Either way, thanks for fixing it!

nwellnhof:
> I can never get a NULL as a result from a getter, but I can use one in a setter (which will be used as an empty string)?

Correct.

PR #35, opened 2015-05-06
Update lua wrapper.

elibarzilay:
I don't know enough lua, and I haven't tried it -- I just noticed some
obvious outdated things, so this is a blind update.

### DIFF
--- a/wrappers/wrapper.lua
+++ b/wrappers/wrapper.lua
@@ -6,28 +6,30 @@ cmark = ffi.load("libcmark")
 
 ffi.cdef[[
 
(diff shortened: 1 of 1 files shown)


jgm:
It needs much more than this -- this dates from an early stage of the API, before we even had the iterator interface.  Probably most of cmark.h can just be copied into the ffi section, but then some of the lua part also needs to be updated.

jgm:
I've updated the luajit wrapper.

elibarzilay:
Yeah, I suspected that it needs more work, I just thought that it would be a good signal to whoever updates it...

PR #36, opened 2015-05-06
New Racket wrapper.

elibarzilay:
Has an almost complete low-level interface, mostly missing the iterator
thing.  Also some higher-level interface: some thin convenient wrappers,
and conversions to/from sexpr format.

### DIFF
--- a/wrappers/wrapper.rkt
+++ b/wrappers/wrapper.rkt
@@ -1,0 +1,188 @@
+#lang racket/base
+
+;; Lowlevel interface
+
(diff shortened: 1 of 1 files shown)


elibarzilay:
BTW, this relies on the NULL PR -- in the conversion of sexprs into cmark nodes, it assumes that it can use the setters with NULLs -- so if that's objectionable (allowing those NULLs) I'll need to somehow update the wrapper to avoid such  sets.

PR #39, opened 2015-05-21
Explicitly add `-fPIC` for static library

gjtorikian:
I'm not very well acquainted with how `cmake` works. Many resources online indicate that `-fPIC` is set by default when building a static library. 

However, I'm finding that's not the case. When trying to natively calls some Houdini methods in https://github.com/gjtorikian/commonmarker/pull/5, the CI runs into [linking issues](https://travis-ci.org/gjtorikian/commonmarker/jobs/63438886). Setting this flag explicitly resolves these issues. 

Perhaps when `COMPILE_FLAGS` is redefined it overrides the default `-fPIC` setting?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -64,7 +64,7 @@ add_compiler_export_flags()
 
 # Disable the PUBLIC declarations when compiling the executable:
 set_target_properties(${PROGRAM} PROPERTIES
-  COMPILE_FLAGS -DCMARK_STATIC_DEFINE)
+  COMPILE_FLAGS "-DCMARK_STATIC_DEFINE -fPIC")
 
 # Check integrity of node structure when compiled as debug:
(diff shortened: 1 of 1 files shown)


gjtorikian:
`CMAKE_C_FLAGS` worked for me.

omasanori:
Debian 8.x (jessie) with CMake 3.0.2 was released, so we can use ``POSITION_INDEPENDENT_CODE`` now, or we should wait until Debian 7.x EOL (2016-04-25)?
(BTW, we have already required 2.8.9, so it should be available AFAIK)

jgm:
It looks like you can conditionalize on cmake version.
http://www.cmake.org/cmake/help/v3.0/variable/CMAKE_VERSION.html
So maybe we could add the feature if the user has a recent enough cmake to support it?

jgm:
Ah, you're right, we are requiring 2.8.9 now.  So, no reason not to just add this.

Issue #44, opened 2015-05-29
UTF8 on windows can lead to Segmentation fault


jeroenooms:
My R bindings are causing [crashes](http://cran.r-project.org/web/checks/check_results_commonmark.html) on windows for certain files containing non-ascii utf8 characters when rendering to xml. It is very difficult to narrow this down because it happens quite randomly and only on windows. R uses mingw-w64 with gcc-4.6.3

One input file that consistently causes the crash is:

```
https://raw.githubusercontent.com/yihui/knitr/1df6eee4ac9387a881db60316c9b334fe21d5133/NEWS.md
```

The strange thing is that there is no particular line that causes the problem. Here it chokes on line 888 but modifying a random line elsewhere in the document can sometimes also fix the problem. Moreover I noticed that enabling `CMARK_OPT_NORMALIZE` will prevent the problem from appearing as well, at least for this particular file.

jgm:
It would be most helpful if you could reproduce the problem
using just cmark (or libcmark).  Then we'd know the problem
is with cmark and not with your interface code.  I couldn't
reproduce any problems converting this file with cmark.

You say the problem is just in rendering to xml, not to
other formats?

What kind of "crashes"?

Are you sure it's non-ascii characters that cause the
problem?  I didn't see any in line 888.

jeroenooms:
As said I had great difficulty diagnosing the exact problem, it is very unpredictable and appears only on windows. It freezes during `cmark_render_xml` (other formats work fine). I'll see if I can build the cmark CLI on windows to have a smaller example. My wrapper is very thin, it just passes along a `cont char*` not too much can go wrong there.

jeroenooms:
Okay I was able to reproduce the problem using a windows build for `cmark.exe`. I put a copy of my build and the file here: http://www.stat.ucla.edu/~jeroen/files/cmark-win32.zip. You will need a windows machine to reproduce this though...

jeroenooms:
I tried building `cmark` it with a more recent vesion of gcc on windows and I keep getting segmentation faults when running this example.

nwellnhof:
This can be reproduced by running

    cmark -t xml NEWS.md

on Windows. I tracked it down to the call to `vsnprintf` in `cmark_strbuf_vprintf`. The Windows version of `snprintf` is not POSIX/C99-compliant as it returns -1 if the output was truncated.

jeroenooms:
Here is some `strace` output: http://pastebin.com/raw.php?i=D6AgsCCd

PR #45, opened 2015-05-29
Cope with broken snprintf on Windows

nwellnhof:
On Windows, snprintf returns -1 if the output was truncated. Fall back to
Windows-specific _scprintf.

Fixes issue #44.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -128,6 +128,7 @@ install(FILES
 # Feature tests
 include(CheckIncludeFile)
 include(CheckCSourceCompiles)
+include(CheckCSourceRuns)
 include(CheckSymbolExists)
 CHECK_INCLUDE_FILE(stdbool.h HAVE_STDBOOL_H)
 CHECK_C_SOURCE_COMPILES(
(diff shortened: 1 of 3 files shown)


jeroenooms:
Thanks! Maybe simply `#ifdef _WIN32` is easier than the cmake route?

nwellnhof:
On 29/05/2015 22:33, Jeroen Ooms wrote:
> Thanks! Maybe simply |#ifdef _WIN32| is easier than the cmake route?

Feature tests are more robust. Testing for `_WIN32` would break older Cygwin 
installations, for example. Also, Microsoft is improving C99 support 
continuously. Maybe they offer a conformant `snprintf` at one point.

jeroenooms:
Okay thanks. Hopefully that fixes it. How did you track down this problem? I've been looking at stack traces all day but the point of crash seemed pretty random.

nwellnhof:
On 29/05/2015 22:43, Jeroen Ooms wrote:
> Okay thanks. Hopefully that fixes it. How did you track down this problem?
> I've been looking at stack traces all day but the point of crash seemed pretty
> random.

I analyzed the stack trace with WinDbg, found that the segfault happened in 
strbuf_puts with the strbuf pointing to an empty string. Then I guessed that 
it might be pointing to the internal OOM value, which proved right. Then I 
added some debug printfs in the places where the OOM value is assigned which 
led to strbuf_vprintf. Having some knowledge of the cmark internals certainly 
helped :)

The whole OOM stuff in strbuf is rather useless since the OOM state isn't 
checked after growing the string buffer, resulting in segfaults which it 
ostensibly tries to protect against. We should rip it out and simply abort on 
every unexpected error from malloc, snprintf, etc. This would have made 
debugging this issue a lot easier.

jeroenooms:
Okay I would have never found this; thanks for fixing it. Hopefully we can merge and tag a release soon and not have to deal with debugging windows internals for a long time to come.

jgm:
Great, thanks!

PR #55, opened 2015-06-07
Abort on strbuf errors

nwellnhof:
Users of the strbuf API are supposed to check for an OOM condition
after appending to strbufs, but:

* This is never done in the whole code base.
* The implementation was flawed because only `ptr` was set to the
  OOM value without adjusting `size` and `asize`. After an error,
  subsequent calls could very well lead to segfaults, contrary to the
  documentation.

Change the code to always abort on errors with a message printed to
stderr. The only alternative is to propagate errors throughout the
whole library which seems infeasible.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -13,11 +13,10 @@
  * assume ptr is non-NULL and zero terminated even for new cmark_strbufs.
  */
 unsigned char cmark_strbuf__initbuf[1];
-unsigned char cmark_strbuf__oom[1];
 
-#define ENSURE_SIZE(b, d)					\
-	if ((d) > b->asize && cmark_strbuf_grow(b, (d)) < 0)	\
-		return -1;
+#define ENSURE_SIZE(b, d)			\
+	if ((d) > b->asize)			\
+		cmark_strbuf_grow(b, (d));	\
 
 #ifndef MIN
 #define MIN(x,y)  ((x<y) ? x : y)
@@ -33,16 +32,13 @@ void cmark_strbuf_init(cmark_strbuf *buf, int initial_size)
 		cmark_strbuf_grow(buf, initial_size);
 }
 
-int cmark_strbuf_try_grow(cmark_strbuf *buf, int target_size, bool mark_oom)
+void cmark_strbuf_grow(cmark_strbuf *buf, int target_size)
 {
 	unsigned char *new_ptr;
 	int new_size;
 
-	if (buf->ptr == cmark_strbuf__oom)
-		return -1;
-
 	if (target_size <= buf->asize)
-		return 0;
+		return;
 
 	if (buf->asize == 0) {
 		new_size = target_size;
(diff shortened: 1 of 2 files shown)


jgm:
I'm trying to understand this better.  In `buffer`, the old `cmark_strbuf_try_grow` is only called by `cmark_strbuf_grow`.  `cmark_strbuf_grow` is called by

- `cmark_strbuf_init`
- `cmark_strbuf_attach` (in the case where the second parameter is NULL), and
- (indirectly via `ENSURE_SIZE`) by `cmark_strbuf_set`, `cmark_strbuf_putc`, `cmark_strbuf_put`, `cmark_strbuf_vprintf`.
- in `houdini_html_u`, by `houdini_unescape_html`.

Only when it is used by `ENSURE_SIZE` does the return value of `grow` get checked and passed on.

I guess my instinct is to be cautious about removing the error reporting behavior, unless it really can't be fixed.  Can you elaborate on the flaw you saw?  Would it be enough to set `size` and `asize` to 0 when `ptr` is set to the OOM value?  Would adding checks for the OOM value to the code base be too complicated?

nwellnhof:
On 07/06/2015 19:46, John MacFarlane wrote:
> I'm trying to understand this better. In |buffer|, the old
> |cmark_strbuf_try_grow| is only called by |cmark_strbuf_grow|.
> |cmark_strbuf_grow| is called by
>
>   * |cmark_strbuf_init|
>   * |cmark_strbuf_attach| (in the case where the second parameter is NULL), and
>   * (indirectly via |ENSURE_SIZE|) by |cmark_strbuf_set|, |cmark_strbuf_putc|,
>     |cmark_strbuf_put|, |cmark_strbuf_vprintf|.
>   * in |houdini_html_u|, by |houdini_unescape_html|.
>
> Only when it is used by |ENSURE_SIZE| does the return value of |grow| get
> checked and passed on.

Correct. But note that the return value of these strbuf functions is never 
checked by any of the users in the rest of libcmark. I changed the return type 
to void without breaking any other code. This means that all errors from 
strbuf are silently ignored.

> I guess my instinct is to be cautious about removing the error reporting
> behavior, unless it really can't be fixed.

I'm not removing error reporting behavior. I'm rather adding it. Before, OOM 
and other errors (like the dreaded vsnprintf thing on Windows) would be 
silently ignored. Now every error will abort immediately with an error message.

> Can you elaborate on the flaw you
> saw? Would it be enough to set |size| and |asize| to 0 when |ptr| is set to
> the OOM value?

Yes, this should be enough.

> Would adding checks for the OOM value to the code base be too
> complicated?

I think it's rather complicated. And the question is what to do when code in 
libcmark gets an OOM error from a strbuf function. Aborting has the same 
effect as my commit. The only alternative is to propagate errors all the way 
down to the public API. This in turn means that every user of the public API 
must check every call to the API for errors. And in most cases, they too have 
no alternative than to abort on an OOM error.

I wouldn't oppose such an approach. It's simply a lot of work without much 
benefit. If someone wants to implement this, I'll happily revert my changes. 
But for now, my commit actually improves error reporting.

Nick

jgm:
OK, fair enough.  We can always revert this if we decide later to add error handling to the API.
This should be a very rare error condition, anyway.

PR #58, opened 2015-06-09
Further optimize utf8proc_valid

nwellnhof:
Assume a multi-byte sequence and rework switch statement into if/else
for another 2% speedup.

### DIFF
--- a/src/utf8.c
+++ b/src/utf8.c
@@ -54,9 +54,11 @@ static int utf8proc_charlen(const uint8_t *str, bufsize_t str_len)
 }
 
 // Validate a single UTF-8 character according to RFC 3629.
+// Assumes a multi-byte UTF-8 sequence.
 static int utf8proc_valid(const uint8_t *str, bufsize_t str_len)
 {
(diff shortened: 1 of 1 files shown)


jgm:
After merging this, I ran my own benchmarks, and I see a slight slowdown (median from 0.250 to 0.255) rather than a speedup.  clang/OSX/64 bit.
I see no change on gcc/linux/32 bit.
I'm happy to keep this change, if you think the code is better.  But I didn't see a speed boost from it.

nwellnhof:
I was testing with gcc 4.9.1, 32-bit Linux, Celeron G1840T @ 2.50GHz (Haswell). I also tested with benchinput.md copied 8 times over because I only get hundredths-of-seconds precision with `time -p`. My results: mean = 1.0750, median = 1.0700 vs mean = 1.1060, median = 1.100

nwellnhof:
I also see a 2% slowdown on clang-602.0.53, OS X 10.10, 64-bit. Feel free to revert the commit.

jgm:
I'll revert.

PR #61, opened 2015-06-25
Changed version variables to functions.

### LINKED ISSUE #60: FFI bindings
Hey guys, I saw that the version strings are external symbols to variables.

Some programming languages support only ffi, as in, all external calls can be made only through functions, variables are not accessible even if they are external, for example C#.

Could we get additional functions to retrieve the version string and integer number?

txdv:
This is easier to access using ffi, since some languages, like C#
like to use only function interfaces for accessing library
functionality.

fixes #60

### DIFF
--- a/api_test/main.c
+++ b/api_test/main.c
@@ -55,8 +55,8 @@ test_continuation_byte(test_batch_runner *runner, const char *utf8);
 static void
 version(test_batch_runner *runner)
 {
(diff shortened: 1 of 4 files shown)


txdv:
I addressed the extern keyword comment.

nwellnhof:
:+1: to merge

jgm:
Thanks!

PR #66, opened 2015-07-16
Split roundtrip_test and leakcheck.

omasanori:
CI status became green even if roundtrip_test failed, since `then` block was not suspended when $? was non-zero. Splitting roundtrip_test and leakcheck addresses this problem.
With this change, builds check ``${TRAVIS_OS_NAME}`` twice, but checking a variable is cheap enough.


jgm:
Thanks - I ran roundtrip_test manually recently and noticed it failed, and was puzzled why travis hadn't reported this.

PR #70, opened 2015-08-06
Prefix utf8proc functions to avoid conflict with existing library

kainjow:
There are linker errors when the library is used with a program that already uses utf8proc, so add prefixes to avoid conflicts.


jgm:
Good idea, thanks.

PR #75, opened 2015-08-10
Fix FileNotFoundError errors on tests when cmark is built from anothe…

kainjow:
…r project via add_subdirectory()

CMAKE_SOURCE_DIR will refer to the top-level source directory, not cmark, so CMAKE_CURRENT_SOURCE_DIR must be used. Also the same for CMAKE_BINARY_DIR. Ideally it should be using the TARGET_FILE CMake generator to get a target's location instead of hard coding the path, but that requires CMake 3.0 or greater.


jgm:
Thanks - I'm not a cmake expert, so this is good to know.

PR #76, opened 2015-08-10
Fix some MSVC warnings

kainjow:
From VS 2013


jgm:
@nwellnhof does this look okay to you?

nwellnhof:
Looks good. I'm not sure about the "Fix MSVC inline errors" commit.

First of all, `static` shouldn't be part of the `CMARK_INLINE` macro. Then, we always compile as C++ under MSVC, so there's no need to use `__inline`. But having an "inline" macro is useful to compile cmark in plain C mode with newer MSVC versions. Is that what you're trying to achieve, @kainjow?

I'd prefer something like:

```
#ifndef CMARK_INLINE
  #if defined(_MSC_VER) && !defined(__cplusplus)
    #define CMARK_INLINE __inline
  #else
    #define CMARK_INLINE inline
  #endif
#endif

...

static CMARK_INLINE void
cmark_chunk_free(cmark_chunk *c) {
```

MSVC 12 and above support many C99 features like mixed declarations and code or `stdbool.h`, so it should be possible to compile cmark as plain C. But inline functions aren't fully supported, so you still have to use `__inline` in C mode. With a `CMARK_INLINE` macro, we could also compile cmark in C mode with some MSVC versions. This is something I planned for a while, so :+1: to merge with the change above.

jgm:
I merged the PR and then added a separate commit, f1b1e2ec71f50a00850db3a68c421a2d79bba115, with @nwellnhof 's recommendations.  @nwellnhof - does this look right?  @kainjow - does this fit your needs?

nwellnhof:
Looks good.

kainjow:
Somehow the api_test CMAKE_C_FLAGS weren't getting applied to that target, so it was getting compiled as C instead of C++, and that is what triggered the error (an error for C clients anyways). I'm not sure why that was the case, as it seems identical to cmark target which compile fine. I tested by compiling the ALL_BUILD target in VS 2013. Very odd. FWIW, I use add_definitions() or target_compile_options() and haven't had issues.

The change looks good. I am running a Jenkins build and will report back.

kainjow:
Looks good.

PR #79, opened 2015-09-02
Better handle libdir

bochecha:
On some OSes we want to install the library in a different location depending on the system architecture.

For example, on a 64 bits Fedora system, the library should be installed in `/usr/lib64`, as opposed to the traditional `/usr/lib`.

This commit allows the builder to specify their own library installation directory, using the variable that the Fedora build process already defines.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,3 +1,8 @@
+if(NOT LIB_INSTALL_DIR)
+  set(LIB_INSTALL_DIR "lib" CACHE STRING
+  "Set the installation directory for libraries." FORCE)
+endif(NOT LIB_INSTALL_DIR)
+
 cmake_minimum_required(VERSION 2.8)
 set(LIBRARY "libcmark")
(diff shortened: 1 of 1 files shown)


bochecha:
I'll note that this is the first time I do something like this with cmake, so I'm not entirely sure I've done the right thing.

Obviously, I'll be happy to improve the patch if needed. :smiley:

bochecha:
So the CI failed, but looking at the results, I'm not sure how I could have broken that with this change.

I mean, I'm not touching the code, only the cmake stuff. :confused:

jgm:
I don't see how this failure could be due to your change.  Probably something having to do with unicode encoding/decoding in the test runner -- and specific to Windows.  I'll have to investigate further on a Windows VM.

jgm:
Thanks for this -- and sorry for the delay.  It couldn't be merged automatically, so I made a separate patch with essentially the change you recommended, crediting you.

Issue #83, opened 2015-09-21
Failed tests on CPython 3.5.0


zmwangx:
Some of the tests are not compatible with CPython 3.5.0. With latest release 0.22.0 (and probably also master, since there's only one, unrelated commit since the latest release), four out of the six `make test` tests fail:

```
	  2 - html_normalization (Failed)
	  3 - spectest_library (Failed)
	  5 - spectest_executable (Failed)
	  6 - smartpuncttest_executable (Failed)
```

A sample test log is here: https://gist.github.com/a13cfb20c900e16a8bf0.

Most of the failures are due to the removed `HTMLParseError` in Python 3.5, and that one is easy to fix. See my proposed changes in #84 . After applying my patch, tests 3, 5, and 6 all pass, but there's still a non-trivial failure in `html_normalization`'s doctest (log is here: https://gist.github.com/12b71e2ef22ef4b9d635):

```python-traceback
Command: "/usr/local/bin/python3" "-m" "doctest" "/tmp/commonmark20150920-37576-vpuo15/cmark-0.22.0/test/normalize.py"
Directory: /tmp/commonmark20150920-37576-vpuo15/cmark-0.22.0/build/testdir
"html_normalization" start time: Sep 20 16:36 PDT
Output:
----------------------------------------------------------
**********************************************************************
File "/tmp/commonmark20150920-37576-vpuo15/cmark-0.22.0/test/normalize.py", line 170, in normalize.normalize_html
Failed example:
    normalize_html("&forall;&amp;&gt;&lt;&quot;")
Expected:
    '\u2200&amp;&gt;&lt;&quot;'
Got:
    '∀&><"'
**********************************************************************
1 items had failures:
   1 of  10 in normalize.normalize_html
***Test Failed*** 1 failures.
```

Looks like somehow the HTML entities have been decoded. I think it is related to the Python 3.5 change to `HTMLParser`: the `convert_charrefs` argument to `__init__`, introduced in Python 3.4, is now default to `True`. But I've tried passing `convert_charrefs=False` to `__init__`; that doesn't immediately fix the problem. I haven't got time to read through `normalize.py`, so perhaps someone else has to look at what's wrong with it.

---

P.S. Quoting from https://docs.python.org/3.5/whatsnew/3.5.html:

[...]

PR #84, opened 2015-09-21
Python 3.5 compatibility fix: add placeholder for HTMLParseError

zmwangx:
See #83 for details of the issue.

`HTMLParseError` was removed in Python 3.5. Since it could never be thrown in Python 3.5+, we simply define a placeholder when `HTMLParseError` cannot be imported.


jgm:
Thanks!

PR #86, opened 2015-10-22
commonmark: fix size_t to int

kainjow:
This fixes an MSVC warning "conversion from 'size_t' to 'int', possible loss of data"


jgm:
Thanks

PR #89, opened 2015-10-29
Correct string length in cmark_parse_document example.

Eljay:
Currently, the example doesn't include the final `*`.


jgm:
Guess I can't count.  Thanks!

PR #92, opened 2015-11-02
Replaced sprintf with snprintf.

mbenelli:


jgm:
We used to use snprintf, but this caused problems on Windows MSVC.  You'll see that the appveyor build failed. Search git history for snprintf and you should find some of the back story.  (I can't remember the full details, but as I recall, MSVC provides a replacement function but it doesn't actually behave the same as snprintf.  Maybe @nwellnhof can elaborate.)

I believe that all of our current uses of sprintf are safe, since we've allocated a buffer big enough to hold the longest possible output in each case.  But if you spot a potentially unsafe use, please bring that to our attention.

nwellnhof:
Yes, older MSVC versions only provide [`_snprintf`](https://msdn.microsoft.com/en-us/library/2ts7cx93.aspx) which has different semantics than C99 snprintf. If we really wanted to use snprintf, we'd have to provide a custom implementation for platforms that don't have native support.

mbenelli:
Sorry for not having checked the history, I just have the need to silence warnings when linking to libcmark; indeed I have left untouched the use of sprintf, as well as strcat and strcopy, in tests. I did a pull request in case it could be of interest for you.
I agree that current use of sprintf are safe, but I don't want to check them on each release.
Thank you very much for your time.

jgm:
There are a couple of ideas [here](http://stackoverflow.com/questions/2915672/snprintf-and-visual-studio-2010), with code, for implementing snprintf for older MSVC versions.  I agree with @mbenelli that it's better to set things up so that we are safe without having to think about each application.

@nwellnhof do you have opinions about any of these proposed replacements?

nwellnhof:
The highest voted answer on StackOverflow looks fine.

jgm:
Great, I've used that code and merged mbenelli's commit.
If someone who has pre-2015 MSVC could test, that would be appreciated!

nwellnhof:
Our Appveyor setup uses a pre-2015 MSVC: https://ci.appveyor.com/project/jgm/cmark/build/1.0.330 (Probably caused by missing include guards in `config.h`.)

PR #103, opened 2016-02-07
blocks: Factorize S_processLines

MathieuDuponchelle:
It's the core of the program and I had too much trouble making
sense of it, two loops with many cases and other code
interspersed hurt my head.

All the tests still passed before rebasing, now I've got the
exact same set of issues as master.


jgm:
Thanks, merged with a few minor changes.
That is better indeed.

PR #104, opened 2016-02-11
blocks: More documentation and refactoring

MathieuDuponchelle:
I finalized the refactoring started in https://github.com/jgm/cmark/pull/103 , some more wouldn't hurt but it's more localized and doesn't really affect my understanding of the control flow in there :) I'll try and see how to offer a generic interface for block parsing extensions now, my test case will be github-style tables

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -1,3 +1,10 @@
+/**
+ * Block parsing implementation.
+ *
+ * For a high-level overview of the block parsing process,
+ * see http://spec.commonmark.org/0.24/#phase-1-block-structure
+ */
+
 #include <stdlib.h>
 #include <assert.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
> Should this say: "returns the last matching node"?

Indeed, fixed

> These names aren't really appropriate -- instead of S_parse_node_item, for example, we should have something like S_parse_node_item_prefix. Similarly for the others.

Makes sense, fixed. FWIW I have also removed the S_ prefix, which I assumed stood for "static". IMHO it's not a useful information at call time, and the `static` keyword is a sufficient indication next to the function definition. Didn't touch the rest of the functions, so the naming is inconsistent in that file, my advice would be to remove the prefixes all around as well, as they're (admittedly not very) confusing.

MathieuDuponchelle:
Hm sorry, didn't note the comments were on the commits themselves, FWIW to keep comments across rebase / commit amendments you may want to comment directly in the "files changed" tab.

PR #108, opened 2016-03-09
Moved the cmake minimum version to top line

tinysun212:
In the file CMakeLists.txt, the required version should be placed to top line. The information could not used at CMake/Modules/CYGWIN.cmake under Cygwin.


jgm:
Thanks, didn't know that.

PR #111, opened 2016-03-18
Add library option to render softbreaks as spaces

PavloKapyshin:
1. Option name was chosen to avoid confusion with “two spaces at the end” feature.

2. `make man/man3/cmark.3` produced extra non-related changes, so I did not do it.

3. There are two places I did not touch regarding this pull request: `src/main.c` and `src/commonmark.c` (I'm not familiar with their code, as I don't use them, so I can't decide how they should behave now).

### DIFF
--- a/api_test/main.c
+++ b/api_test/main.c
@@ -788,6 +788,11 @@ static void line_endings(test_batch_runner *runner) {
   STR_EQ(runner, html, "<p>line<br />\nline</p>\n",
          "crlf endings with CMARK_OPT_HARDBREAKS");
   free(html);
+  html = cmark_markdown_to_html(crlf_lines, sizeof(crlf_lines) - 1,
+                                CMARK_OPT_DEFAULT | CMARK_OPT_NOBREAKS);
+  STR_EQ(runner, html, "<p>line line</p>\n",
+         "crlf endings with CMARK_OPT_NOBREAKS");
+  free(html);
 
   static const char no_line_ending[] = "```\nline\n```";
   html = cmark_markdown_to_html(no_line_ending, sizeof(no_line_ending) - 1,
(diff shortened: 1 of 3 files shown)


jgm:
The man, commonmark, and latex renderers all provide a "column width" argument (set by `--width` in the command-line tool).  If this is 0, then all soft breaks are rendered as newlines.  Otherwise, soft breaks are treated as spaces, and cmark inserts newlines where appropriate to ensure that the lines are as close as possible to the specified width.  If you set width to a very high number, this will mean that all soft breaks are spaces.

The reason I didn't add this feature to the HTML renderer is that it's pretty expensive (performance-wise).  For the HTML renderer we focus on speed.  Your change wouldn't be expensive, but it has the disadvantage of providing a non-uniform way of setting soft break behavior (you'd do it one way for HTML and a different way for the others).

Not really sure what's best here; these are my thoughts.

PavloKapyshin:
There is functional non-uniformity between HTML renderers in two reference implementations (commonmark.js has functionality provided by this patch), so, from the practical standpoint (similarly to not implementing HTML wrapping which is better handled by specialized tools), it might be ok to avoid forcing users to implement such trivial thing (similarly to providing simple API for CommonMark-to-HTML) with hacks like a) using input -> (html hardbreaks=1) -> re.sub(br, " ") or b) using input -> (commonmark width=0 hardbreaks=1) -> html.

jgm:
Yes, that's a good point.

+++ Pavlo Kapyshin [Mar 21 16 03:43 ]:

jgm:
One question is whether this option should apply to all renderers (currently it only affects HTML, which should at least be documented).

PavloKapyshin:
I added corresponding note.

jgm:
I'm going to make a subsequent commit  that implements the option for all renderers, and adds a `--nobreaks` command line option.

PR #120, opened 2016-03-28
Add first regression tests

nwellnhof:
I think it's a good idea to add regression tests after fixing bugs. In most cases, this is really easy using the spec test infrastructure.


jgm:
Good idea.

Issue #100, opened 2016-02-04
Extension support in libcmark


MathieuDuponchelle:
Hello, I always see "extensions" mentioned on discussions of features in CommonMark (for example the discussion about tables).

Does libcmark itself support *actual* extensions, and if so is there any guide on how to implement one, and an index of common extensions, or are extensions purely conceptual extensions to the specification, up to individual implementations to add ?

I'm pretty sure I could just read the code and find out but

* I'm lazy
* This question may be useful for someone else wondering the same thing.

Thanks for all your work on CommonMark / pandoc!

jgm:
libcmark currently has no mechanism for extensions.

MathieuDuponchelle:
Are there any plans for this ?

jgm:
No definite plans at the moment.
Of course you can do quite a lot of customization already by
walking the AST with the iterator interface.  See jgm/lcmark
for a system that makes this easy to do with lua filters.

MathieuDuponchelle:
Interesting, thanks, however afaiu some extensions will need to define rules at the parsing stage? ie for example the table extension would need to define new node types, and set state such as IN_TABLE when the beginning of a table is encountered?

MathieuDuponchelle:
If that was not clear, I like tables :)

MathieuDuponchelle:
By the way if you're interested in documenting libcmark's API with CommonMark using the tool I'm developing (here's a demo video of it I made two weeks ago -> https://www.youtube.com/watch?v=jQd1zJIJQRE) I can try and give it a go :)

jgm:
>   Interesting, thanks, however afaiu some extensions will need to define
>   rules at the parsing stage? ie for example the table extension would
>   need to define new node types, and set state such as IN_TABLE when the
>   beginning of a table is encountered?

Yes, that's much harder to accomplish.  (Though currently
you could do it in a sort of hackish way, for example, by
including the table in a processing instruction

    <?table
    | my | table |
    |----|-------|
    | 2  | 3     |
    ?>

which a filter can intercept, process, and replace with raw
HTML or whatever is the target format.)

The javascript library markdown-it contains a flexible
extension mechanism; you might look at it.

MathieuDuponchelle:
> The javascript library markdown-it contains a flexible extension mechanism; you might look at it.

I kind of want to use libcmark itself as it's hell of fast, also for my customization needs I currently use commonmark-py , as my tool is written in python

MathieuDuponchelle:
[...]

PR #123, opened 2016-04-26
Implement parsing extensions

MathieuDuponchelle:
Reference discussion: https://github.com/jgm/cmark/issues/100

I just rebased the branch on master and tested it, make passes with no new warnings all along the branch, valgrinding a table test shows no leaks.

I applied the initial review comments from @jgm and @nwellnhof , which were to not expose cmark_strbuf, which is thus included in the sources of the reference "core extension", and to not expose API using cmark's bool type, which is replaced by integers, rendering installation of config.h unnecessary.

The design implemented is the second one I proposed in the above discussion, here's a brief summary:
- A new structure is exposed, `cmark_syntax_extension`, which exposes hooks that match the block and inline parsing strategies outlined in http://spec.commonmark.org/0.25/#appendix-a-parsing-strategy .
- This proposal does not cover the implementation of new block types, instead syntax extensions need to create blocks with types defined by cmark itself, including `CMARK_NODE_CUSTOM_BLOCK` and `CMARK_NODE_CUSTOM_INLINE`, this already covers a wide range of use cases.
  
  That is because in my opinion, we want cmark to guarantee that the output AST will always be correctly structured (with the exception of custom inlines and blocks), with no "impossible" or "maningless" block nesting situations.
  
  An interesting enhancement would be to refactor the custom inline and block types to let extensions provide virtual methods such as "can_contain" for them, but this will be somehow fragile as "inter-extension negotiation" will not be possible.
- An important part of this proposal is that cmark implement and expose block types for which it does not yet have a formal syntax specification, such as tables or strikethroughs. I believe this is desirable because this will allow multiple syntax extensions to compete for a given desired output, without tying cmark's specification to a specific syntax, while still making sure the produced output is correct. It makes sense when looking at it as a testing ground if you will.

It is important to state that this design doesn't prevent us from implementing negotiation in the future if we so desire.

I think this branch is now approaching what I'd consider mergeable quality, with the exception of the plugin code, which I think is mostly correct on Linux, but will not work on Windows. If @nwellnhof is interested in tackling the abstraction for Windows, then all the best, otherwise the code can be merged without plugin loading / discovery for now, as it is already useful for people using cmark as a library (such as I).

Last thing, this last rebase round has been pretty involved, and I'd very much like (if we all agree that extensions are desirable) that merging this branch be made a priority, I will of course address reviews in a timely manner.


MathieuDuponchelle:
yay MSVC \o/ I'll have a look at the issues tomorrow, hoping it does not just uncover a new layer :)

src/cmark.h
```diff
@@ -592,6 +592,29 @@ int cmark_parser_has_partially_consumed_tab(cmark_parser *parser);
...
+ *
+ * If 'parent' isn't the kind of node that can accept this child,
+ * this function will back up till it hits a node that can, closing
+ * blocks as appropriate.
+ */
+CMARK_EXPORT
+cmark_node*cmark_parser_add_child(cmark_parser *parser,
+                                  cmark_node *parent,
+                                  cmark_node_type block_type,
+                                  int start_column);
```
nwellnhof: Can't extensions simply use `cmark_node_append_child`?

src/cmark.h
```diff
@@ -592,6 +592,29 @@ int cmark_parser_has_partially_consumed_tab(cmark_parser *parser);
...
+/** Advance the 'offset' of the parser in the current line.
+ *
+ * See the documentation of cmark_parser_get_offset() and
+ * cmark_parser_get_column() for more information.
+ */
+CMARK_EXPORT
+void cmark_parser_advance_offset(cmark_parser *parser,
+                                 const char *input,
+                                 int count,
+                                 int columns);
```
nwellnhof: I think the `input` chunk should be part of `cmark_parser`, but that's an issue with the parser code.

src/node.c
```diff
@@ -150,6 +155,15 @@ cmark_node_type cmark_node_get_type(cmark_node *node) {
...
 }
 
+int cmark_node_set_type(cmark_node * node, cmark_node_type type) {
+  if (type == node->type)
+    return true;
+
+  free_node_as(node);
+  node->type = type;
+  return true;
+}
```
nwellnhof: I can see why extensions need this function, but in its current form, usage should be restricted to extension authors who know what they're doing. A safe public function should

- check whether the new node type is valid for the node's parent.
- initialize the node to its default state (unless the content can be reused).

Maybe the parser could first assign an internal node type like UNKNOWN_BLOCK which can later be changed to a known block type?

src/cmark.h
```diff
@@ -86,6 +86,8 @@ typedef struct cmark_node cmark_node;
 typedef struct cmark_parser cmark_parser;
 typedef struct cmark_iter cmark_iter;
 
+typedef void (*CMarkNodeUserDataFreeFunc) (void *user_data);
```
nwellnhof: Can you please stick to snake case? A short name like `cmark_free_func` would be OK with me.

src/node.h
```diff
@@ -55,6 +55,7 @@ struct cmark_node {
   struct cmark_node *last_child;
 
   void *user_data;
+  CMarkNodeUserDataFreeFunc user_data_free_func;
```
nwellnhof: Would it be possible to only add this field to node types that actually need it?

src/cmark.h
```diff
@@ -264,6 +264,16 @@ CMARK_EXPORT const char *cmark_node_get_literal(cmark_node *node);
...
 
+/** Return the string content for all types of 'node'.
+ *  The pointer stays valid as long as 'node' isn't freed.
+ */
+CMARK_EXPORT const char *cmark_node_get_string_content(cmark_node *node);
+
+/** Set the string 'content' for all types of 'node'.
+ *  Copies 'content'.
+ */
+CMARK_EXPORT int cmark_node_set_string_content(cmark_node *node, const char *content);
```
nwellnhof: The problem with `set_string_content` is that children of a block node may have pointers ("chunks") into its string content. So you can't simply call this function on a paragraph node, for example.

src/cmark.h
```diff
@@ -333,6 +333,15 @@ CMARK_EXPORT const char *cmark_node_get_fence_info(cmark_node *node);
...
 
+/** Sets code blocks fencing details
+ */
+CMARK_EXPORT int cmark_node_set_fenced(cmark_node * node, int fenced,
+    int length, int offset, char character);
+
+/** Returns code blocks fencing details
+ */
+CMARK_EXPORT int cmark_node_get_fenced(cmark_node *node, int *length, int *offset, char *character);
+
```
nwellnhof: This commit doesn't seem related to extensions.

src/blocks.c
```diff
@@ -973,9 +1005,29 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
...
+      cmark_node *new_container = NULL;
+
+      for (tmp = parser->syntax_extensions; tmp; tmp=tmp->next) {
+        cmark_syntax_extension *ext = (cmark_syntax_extension *) tmp->data;
+
+        if (ext->try_opening_block) {
+          const char *input_cstr = cmark_chunk_to_cstr(input);
+
+          new_container = ext->try_opening_block(
+              ext, indented, parser, *container, input_cstr);
```
nwellnhof: It would be more efficient to pass a `char*` and size pair instead of converting to a C string.

src/cmark.h
```diff
@@ -688,6 +726,154 @@ void cmark_parser_advance_offset(cmark_parser *parser,
...
+ * For now the extensibility of cmark is not complete, as
+ * it only offers API to hook into the block parsing phase
+ * (<http://spec.commonmark.org/0.24/#phase-1-block-structure>).
+ *
+ * See 'cmark_plugin_register_syntax_extension' for more information.
+ */
+
+/** The prototype plugins' init function should follow.
+ */
+typedef int (*PluginInitFunc)(cmark_plugin *plugin);
```
nwellnhof: Camel case and `cmark_` prefix, please. Also for the other function types below.

src/node.h
```diff
@@ -69,6 +69,8 @@ struct cmark_node {
 
   cmark_strbuf string_content;
 
+  cmark_syntax_extension *extension;
```
nwellnhof: Would it be possible to add this field only to node types that need it? "Custom" or "extension" nodes, for example.

src/cmark.h
```diff
@@ -688,6 +726,154 @@ void cmark_parser_advance_offset(cmark_parser *parser,
...
+ * It is the responsibility of the parser to create and add the
+ * new block with cmark_parser_make_block and cmark_parser_add_child.
+ * If 'try_opening_block' is NULL, the extension will have
+ * no effect at all on the final AST.
+ */
+struct cmark_syntax_extension {
+  MatchBlockFunc          last_block_matches;
+  OpenBlockFunc           try_opening_block;
+  char                  * name;
+};
```
nwellnhof: The struct definition should be in a private header.

src/parser.h
```diff
@@ -43,6 +43,7 @@ struct cmark_parser {
   /* Options set by the user, see the Options section in cmark.h */
   int options;
   bool last_buffer_ended_with_cr;
+  cmark_llist *syntax_extensions;
```
nwellnhof: I'd add a field `struct cmark_syntax_extension *next` directly to `struct cmark_syntax_extension` and change this field to point to the first extension. Then there's no need for `cmark_llist` at all.

src/blocks.c
```diff
@@ -1266,7 +1293,13 @@ cmark_node *cmark_parser_finish(cmark_parser *parser) {
...
+
+  res = parser->root;
+  parser->root = NULL;
+
+  cmark_parser_reset(parser);
+
+  return res;
 }
 
 int cmark_parser_get_line_number(cmark_parser *parser) {
```
nwellnhof: This commit doesn't seem to be related to extensions.

src/cmark.h
```diff
@@ -558,6 +558,9 @@ void cmark_parser_free(cmark_parser *parser);
 CMARK_EXPORT
 void cmark_parser_feed(cmark_parser *parser, const char *buffer, size_t len);
 
+CMARK_EXPORT
+void cmark_parser_feed_reentrant(cmark_parser *parser, const char *buffer, size_t len);
+
 /** Finish parsing and return a pointer to a tree of nodes.
  */
 CMARK_EXPORT
```
nwellnhof: This commit doesn't seem to be related to extensions.

nwellnhof:
Some general notes:

> This proposal does not cover the implementation of new block types, instead syntax extensions need to create blocks with types defined by cmark itself, including CMARK_NODE_CUSTOM_BLOCK and CMARK_NODE_CUSTOM_INLINE, this already covers a wide range of use cases.

But that's not what your table and strikethrough extensions do. They define their own block types in the cmark core. If your extension nodes used the custom node types, it would be more convincing that this works for other extensions, too.

Regarding dynamic loading and plugin discovery: I think we should leave this out for now, and only have a function to register an extension with a parser.

It might also be a good idea to move everything related to extensions to a separate public header file. There are some functions like `cmark_set_node_type` or `cmark_set_string_content` that are unsafe in their current form and should only be used by extension authors. If we put such functions into `cmark.h`, people will start using them and complain if things start to break. The extension header file should come with a warning that this is a semi-public API only for extension authors.

src/cmark.h
MathieuDuponchelle: No, I suggest you look at the implementation of `add_child` in `src/blocks.c` , this function finalizes open parent blocks, and creates the new block as open. This is really a parsing-specific method, and it cannot be replaced by `cmark_node_append_child`

src/cmark.h
MathieuDuponchelle: ack

src/node.c
MathieuDuponchelle: > check whether the new node type is valid for the node's parent.

Indeed, that would be a valid case for returning false in there, will do

> initialize the node to its default state (unless the content can be reused).

Note that this function does (re)initialize the node to its initial type, that's what "free_node_as" is doing, in case you're not aware node types are implemented as a union.

> Maybe the parser could first assign an internal node type like UNKNOWN_BLOCK which can later be changed to a known block type?

I'm not sure why we'd want to do that? Note that the intended usage for this function is also to modify a previously defined node type, as explained in the message for the commit introducing this API (https://github.com/MathieuDuponchelle/cmark/commits/extensions)

src/cmark.h
MathieuDuponchelle: As you like, I'm used to camel case for such types but if that's not the convention I'll change it

src/node.h
MathieuDuponchelle: Obviously not :)

src/cmark.h
MathieuDuponchelle: Indeed, but manipulating it is legit at parsing time, once again see `src/blocks.c` for examples of that

src/cmark.h
MathieuDuponchelle: Not directly indeed. Wouldn't mind getting review and merging it anyway, it's not such a huge change :)

src/blocks.c
MathieuDuponchelle: True, I have purposefully not tried to optimize that code, as I want it to be as clear as possible at least for initial review, but this is an exposed interface so I agree this is in scope, I'll have a look

src/cmark.h
MathieuDuponchelle: Indeed

src/node.h
MathieuDuponchelle: Nope, an extension might (and some of mine do) create a regular code block for a different syntax for example, cmark needs to know who to ask if they can be closed.

MathieuDuponchelle:
Thanks for the review, answering the general comment here:

> But that's not what your table and strikethrough extensions do. They define their own block types in the cmark core. If your extension nodes used the custom node types, it would be more convincing that this works for other extensions, too.

That is what my extensions do, they don't define their own block types, cmark does, it so happens that the commits for adding some of them live in the same branch. For reference here's the code for a bunch of other extensions:

https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_gtkdoc_extension.c
https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_include_scanner.c

If you look at https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_gtkdoc_extension.c#L256 for example, you will see that this extension instantiates a regular CMARK_NODE_CODE_BLOCK, it's a totally legit use case for the *syntax* extensions design I propose to create standard block types according to their own syntax rules, this makes them robust (block precedence is enforced by cmark's core) and concise (rendering for all possible output formats is handled by cmark's core)

The key element in this proposed design is that extensions do not define their own block types, as I believe there is no satisfying way to ensure inter-extension block precedence / can_contain rules. That is also why my extensions do not use the "CMARK_NODE_CUSTOM_*" types, as I am not interested in a fragile solution, or in implementing format-specific rendering code. There is also no actual correlation between the "convicingness" (is that a word?) of my proposal and the fact that my extensions (do not) use these types.

Note also that the two new block types I implemented in cmark are two extensions for which there is (or at least seems to be) a clear support intention for upstream in the future. I think it can't really hurt to implement the rendering code for them ahead of definitive syntax rules, quite the contrary actually.

> Regarding dynamic loading and plugin discovery: I think we should leave this out for now, and only have a function to register an extension with a parser.

ack, I'll split that out

> It might also be a good idea to move everything related to extensions to a separate public header file. There are some functions like cmark_set_node_type or cmark_set_string_content that are unsafe in their current form and should only be used by extension authors. If we put such functions into cmark.h, people will start using them and complain if things start to break. The extension header file should come with a warning that this is a semi-public API only for extension authors.

That could make sense, along with an "unstable" warning, and maybe a preprocessor directive to enable it (CMARK_ENABLE_EXTENSION_API or something like that)

I would be interested in @jgm's opinion on that last point

Last thing, @nwellnhof I noticed you made a few comments on outdated diffs, and intentionally didn't address these, can you make these comments again on the current diff?

I've had a lot of work today, and what little time I had to work on cmark I spent addressing this (very appreciated review), I will make the agreed-upon changes tomorrow (I hope :)

Thanks!

src/cmark.h
MathieuDuponchelle: It isn't directly, though I can't imagine another use case for it. Here's why I need it for reference:

https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_include_extension.c#L80

This allows me to offer a context-aware / recursive inclusion syntax, that a preprocessor could not. It was how I implemented it a first but things like:

```
\```
{{ do_not_try_to_include_this_really_we_are_in_a_code_block_stop_embarassing_yourself }}
\```
```

could potentially break. By the way formatting this example makes me wonder how we're supposed to escape the code block open / code block close patterns, but that's a completely different issue :)

src/node.c
MathieuDuponchelle: Checking the parent type is done

src/cmark.h
MathieuDuponchelle: Done

src/cmark.h
MathieuDuponchelle: Done

src/parser.h
MathieuDuponchelle: I'd rather have a generic container than intrusive linked lists where one has to reimplement standard iterating and freeing functions.

src/blocks.c
MathieuDuponchelle: It isn't, but it's valuable anyway

src/cmark.h
MathieuDuponchelle: Moved this to a separate public header

src/cmark.h
MathieuDuponchelle: Moved this to a separate public header

src/cmark.h
MathieuDuponchelle: Done

src/node.c
MathieuDuponchelle: Moved this to a separate public header

src/parser.h
MathieuDuponchelle: Besides, this generic structure is also used to specify "special inline chars" later in that branch

MathieuDuponchelle:
I apologize for the noise here (though github should apologize really), I stick to a rebase workflow because I prefer having a clean history to work with, instead of piling up commits on top of my branch.

src/blocks.c
MathieuDuponchelle: Done

MathieuDuponchelle:
I have addressed all the review comments, either correcting the issues that were pointed out or answering the comments.

This branch now installs a new public header "cmark_extension_api.h", separate from cmark.h, to be included and used by syntax extension writers.

Additionally, I've also separated the plugin loading and discovery code to its own commit, however after thinking back on it I'd really like @nwellnhof to come true on the windows port promise: this is a pretty important part of the functionality that you're asking me to take away from the branch, and you have pretty vehemently protested my proposal of using the GLib, which would have offered us a portable and well-tested plugin interface.

It would be a bit ugly to oppose this proposal, state that you can port the code to Windows, then just ask me to drop it when comes the time to actually do what you promised. 

If you do not have time to work on that, I propose we merge the linux implementation to start with.

@jgm, do you have time for reviewing this branch these days?

jgm:
+++ Mathieu Duponchelle [Apr 27 16 16:08 ]:
>   [2]@jgm, do you have time for reviewing this branch these days?

It's crunch time here, but things should open up in another
week or two, I think.

MathieuDuponchelle:
@jgm , ack, can you prioritize reviewing and merging this if you're happy with the design over merging (significant) patches in cmark? Keeping this branch cleanly rebased on master does require a bit of extra work every now and then and I also have more than enough on my plate already as well :)

nwellnhof:
I just think you have a better chance at getting anything merged if you try to keep the initial pull request as small as possible. So

* I'd leave out any commits that aren't related to extensions. These should be discussed separately.
* I'd start with block extensions only (or inline only).
* I'd leave out plugin discovery for now. This can and should be discussed separately.
* Also, I still don't think that it's a good idea to have the code for "core" extensions in the cmark repo, at least for now.

> If you look at https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_gtkdoc_extension.c#L256 for example, you will see that this extension instantiates a regular CMARK_NODE_CODE_BLOCK, it's a totally legit use case for the syntax extensions design I propose to create standard block types according to their own syntax rules, this makes them robust (block precedence is enforced by cmark's core) and concise (rendering for all possible output formats is handled by cmark's core)

I didn't say that extensions must use custom node types. If you just need a different syntax for "standard" nodes, that's fine. But relying on hardcoded extension node types isn't very extensible. What if someone wants to write an extension for a new node type that none of the existing types map to? And your solution for tables is even worse. You actually include struct definitions for your particular idea of table nodes in `node.h`. This is half-baked at best. It wouldn't allow any competing table extensions to add additional fields.

I also don't like that two additional pointers are added to `struct cmark_node`. That's 16 bytes on 64-bit systems that everyone has to pay for, whether they use extensions or not. I haven't given it much thought, but IMO, extension nodes should be implemented as follows:

Introduce a new node type `CMARK_EXT_NODE`. This node type is allowed anywhere and can have any kind of children. (We could allow more control about where extension blocks can be placed later, maybe starting with separating block and inline extension nodes. I don't think this is a priority for now.) This node type has the following struct embedded in the `as` union:

    struct cmark_ext_node {
        struct cmark_ext_node_type *ext_type;
        void *ext_data;
    };

`ext_data` points to extension-specific node data. `cmark_ext_node_type` contains information to identify extension nodes and a couple of functions pointers for node-specific operations (kind of a "class" metaobject):

    struct cmark_ext_node_type {
        cmark_syntax_extension *extension;
        char *name;
        void (*free)(cmark_node *node);
        // Possibly some functions for HTML rendering
        void (*start_html)(cmark_node *node, cmark_strbuf *html);
        void (*end_html)(cmark_node *node, cmark_strbuf *html);
    };

This doesn't require to bake any knowledge about extensions into the cmark core, should be truly extensible, and doesn't increase memory requirements for standard nodes. It also makes it possible to have custom renderers for extension nodes.

MathieuDuponchelle:
@nwellnhof to be honest it feels like you're mostly paying attention to the form of this proposal, and not at all to its actual design. I can break up the branch, but it makes no sense if we don't agree on the design first.

> I haven't given it much thought, but IMO, extension nodes should be implemented as follows

I think this is the issue here. Your proposed design is exactly what I first implemented, see https://github.com/MathieuDuponchelle/cmark/commit/a683b597c437df30370f014d18b6a903d6a9e942

I judged this solution inferior (and still do) after giving it a bit more thought, because of this issue, which you don't address at all in your proposal:

**How can extensions implemented this way negotiate what block types the block types they define can contain and cannot be contained by, including block types provided by other extensions?**

Short answer is: your proposed design cannot ensure this, and saying "these new block types can contain and be contained by anything" is just not satisfying.

Regarding the "half-bakedness" of my table proposal, your criticism is that "extensions cannot add additional fields" . I'm not sure what exactly you mean by "additional fields", but the same could be said of other block types, however I have found no trouble in providing extra syntax rules for them.

Before going any further in that discussion, I'll wait for your answer to the issue I exposed above.

MathieuDuponchelle:
By the way just clearing things out:

> I also don't like that two additional pointers are added to struct cmark_node

This is not the case, only one pointer is added to support extensions, the pointer to the extension that created the node, surely that's not such a huge price to pay for extensibility.

The other pointer is the pointer to a free function for the already existing user_data field, which IMHO should have been there in the first place.

nwellnhof:
> I can break up the branch, but it makes no sense if we don't agree on the design first.

Exactly. That's why we should discuss the design first, then find a way to break up the work into separate pieces which can be reviewed and merged one by one.

> How can extensions implemented this way negotiate what block types the block types they define can contain and cannot be contained by, including block types provided by other extensions?

> Short answer is: your proposed design cannot ensure this, and saying "these new block types can contain and be contained by anything" is just not satisfying.

If you allow extension nodes to tell that they should act like a certain core node type in a certain context, it's very well possible to make extensions work together, even if they don't know anything about each other. For example, a table node could specify that it can appear wherever a paragraph can, and a table cell could say that it can contain whatever a document can contain. Here's a sketch of how this could be implemented: https://gist.github.com/nwellnhof/89077f57e8da0cf13631bf4a4a62ba9f

This should work for all practical use cases. It wouldn't be possible to add, say, a table row from table-extension-A to a table from table-extension-B. For that, you'd really need a centrally defined notion of what a table, table row, or table cell is. But I can't imagine where something like this would be needed. Even if it turns out that we need some well-defined extension types in the cmark core, it wouldn't rule out my proposal.

But IMO, this is a rather unimportant feature that could even be left out completely. What's the worst that could happen? If someone adds a table cell to a node that isn't a table row, they'll only get invalid HTML from the renderer. It would be nice to detect cases like this, but I don't think it's crucial.

> The other pointer is the pointer to a free function for the already existing user_data field, which IMHO should have been there in the first place.

The existing `user_data` field is used by external code which is typically the code that calls `cmark_node_free` and actually knows about the lifetime of nodes. The main motivation for the user data field was to simplify memory management for the Ruby and Perl bindings, but it could be used by C client code as well. This means that with your approach, you'll need a *third* field if you want to allow extra data for extensions. The `free_func` and extension pointer could be combined in a single type object like in my proposal, but the important point is that extensions should add as little overhead as possible if they're not used at all.

MathieuDuponchelle:
> For example, a table node could specify that it can appear wherever a paragraph can, 
and a table cell could say that it can contain whatever a document can contain.

This is just terrible design.

> What's the worst that could happen? If someone adds a table cell to a node that isn't a table row, they'll only get invalid HTML from the renderer.

Well apart from a crash that's indeed the worst thing I think cmark could do, I'm not sure how you can discard it as benign tbh?

> The existing user_data field is used by external code which is typically the code that calls cmark_node_free and actually knows about the lifetime of nodes.

I'm not sure why you make assumptions about usage of what seems like a general purpose API to me. Furthermore, there is a pretty bad flaw with the limited use case you're presenting, which is that there either needs to be only one user for this user_data, or that each user has to keep track of the nodes which user_data it has to free.

> The main motivation for the user data field was to simplify memory management for the Ruby and Perl bindings, but it could be used by C client code as well. This means that with your approach, you'll need a third field if you want to allow extra data for extensions.

So you're telling me this "user_data" is actually really a garbage collection helper, not meant for anything else?

> The free_func and extension pointer could be combined in a single type object like in my proposal, but the important point is that extensions should add as little overhead as possible if they're not used at all.

I'm not interested in your proposal to be honest, I think I'll wait for @jgm 's input now.

MathieuDuponchelle:
By the way a better API for user data would be equivalent to <https://developer.gnome.org/gobject/unstable/gobject-The-Base-Object-Type.html#g-object-set-data> and <https://developer.gnome.org/gobject/unstable/gobject-The-Base-Object-Type.html#g-object-set-data-full> , with nodes maintaining a reference map, this would make it actually robust, that's a more intrusive change and requires a hash table implementation though.

MathieuDuponchelle:
I'll be in holidays for the next two weeks, going to <https://wiki.gnome.org/Hackfests/GstSpringHackfest2016> , I'll check on that issue from time to time but I can't promise immediate reactivity :)

MathieuDuponchelle:
@jgm , ping?

MathieuDuponchelle:
Hello @jgm, I just rebased this branch on master, and this time it was a very painful experience. I'd really appreciate it if I could get some feedback :)

jgm:
I have tried recently to absorb your changes, but got stuck
on rebasing, so thank you for doing that. You've dumped a
huge number of changes into this PR, and I haven't had a
correspondingly huge amount of free time.  I'm afraid I may
be in touch only intermittently over the next two weeks.
After that I should have more time.

Do you have a narrative describing at a high level how the
proposed extension system works?

I may misunderstand things, but at my present incomplete level
of comprehension of your approach, I have some worries.
For example, it seems that extensions need to add to
cmark_node_type in cmark.h (71a4105, 2e8be01).  Extensions
will need to be more decoupled from the core than that,
somehow.

I'd also hope for a way to do things that exposes fewer of
the nitty-gritty details of the parser in the public API
(things like cmark_parser_has_partially_consumed_tab).

Sorry, these impressionistic comments are all I have at the
moment.  I understand your frustration; I'll try to take a
closer look in the next week if I can.

MathieuDuponchelle:
@jgm, the idea with this branch is that I would like cmark, given an AST, to always produce a correct output, except if "custom nodes" are used.

The first solution I worked on in #100 couldn't meet this requirement, as it expected extensions to define new block types, and I could not see any solution to let the "can contain / can be contained by" rules work between block types defined by multiple extensions.

I have implemented multiple extensions using my (not up to date and bundled) fork in hotdoc, for example https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_gtkdoc_extension.c , and none of them actually needed to define new types, for example the gtk-doc syntax handles this syntax:

```
I am linking to this_function()
```

The new node types I have defined in cmark are types which we will want in the future, for which we haven't yet specified a syntax, and I figured it couldn't hurt to implement rendering support for them "preemptively".

An advantage of that solution is that extensions that wish to propose new syntax ruies for existing node types do not need to define duplicate node types, their containing rules and rendering methods for each of the format commonmark currently supports. This could be solved at application level by post processing the AST but that just isn't very elegant.

It's also unclear to me how much of the internals of each renderer we would need to expose (see for example "plain" in the html renderer).

It should be noted that extensions *can* define their own custom types already, through CUSTOM_* nodes, and if their API was slightly reworked (with on_start and on_end made into function pointers), they would be pretty adequate for users who do not consider correct output an absolute requirement. I considered this was out of scope for this proposal, and left it out.

So to answer your concern, extensions **are** decoupled from the core :)

Regarding the exposed implementation details, if we want extensions to participate in the parsing process, they do need to be exposed. I haven't yet used "partially_consumed_tab" in any of my extensions, but considered it was worth exposing, as I imagine it can inform the decisions made by a syntax extension.

As @nwellnhof suggested, I exposed this new API in a separate header, which we could advertise as unstable at the beginning, possibly by #ifndefing its contents with CMARK_UNSTABLE_API ?

MathieuDuponchelle:
Fixed the leaks, @jgm note that the first commit in that branch should be merged ASAP if correct, as it prevents adding user-created nodes to the AST.

src/node.c
```diff
@@ -30,9 +30,6 @@ static bool S_can_contain(cmark_node *node, cmark_node *child) {
   if (node == NULL || child == NULL) {
     return false;
   }
-  if (NODE_MEM(node) != NODE_MEM(child)) {
```
jgm: This is there for a reason.  @vmg commented on his pull request: "Every single node is now also aware of its allocator, which means that we can choose an allocator when creating new nodes and check that you cannot insert a node from a specific allocator into a document tree created by another allocator."  @vmg and @nwellnhof can perhaps comment further on why this is needed.  It shouldn't prevent adding nodes created using `cmark_node_new` to the AST; you just need to make sure that the new node is created using the same memory allocator as the original document, right?

vmg: That's right. If you use `cmark_node_new`, you can only add that node to a document created from `cmark_parser_new` (since these APIs use the default system allocator)... Mixing and matching nodes is not a good idea. A library user should always use the same allocator, hence this check is important!

MathieuDuponchelle: @vmg , I hadn't noticed this new *_with_mem API. I'm not sure I see the use case, but anyway the fact that "a library user should use the same allocator" should be documented, I couldn't find any mention of that in cmark.h :)

I figured out what the actual issue was here, and am not sure how to best solve it:

The "core extensions" are not statically linked with the executable, but dlopened, however we are passing -fvisibility=hidden to the compiler, and DEFAULT_MEM_ALLOCATOR ends up having a different address in cmark_node_new when called from the dlopened library. That's a pretty bad problem, especially as there is no API to retrieve the cmark_mem associated to a given parser, I can (and will) work around that in the core extensions because I #include private header files, including "parser.h", however externally defined extensions will not have that chance.

How do you think we should go about that?

MathieuDuponchelle: I just pushed the fixed version, which calls:

```
res = cmark_node_new_with_mem(CMARK_NODE_TEXT, parser->mem);
```

however as I explained this will not be an option for externally-defined extensions.

nwellnhof: On 09/06/2016 00:02, Mathieu Duponchelle wrote:
> I figured out what the actual issue was here, and am not sure how to best
> solve it:
>
> The "core extensions" are not statically linked with the executable, but
> dlopened, however we are passing -fvisibility=hidden to the compiler, and
> DEFAULT_MEM_ALLOCATOR ends up having a different address in cmark_node_new
> when called from the dlopened library.

You should probably call `dlopen` with the `RTLD_GLOBAL` flag.

jgm: @vmg could you say more about why it would be bad to have nodes in the same tree that have different allocators?

MathieuDuponchelle: @nwellnhof nope that is not the purpose of this flag but nice try, I thought about it too

MathieuDuponchelle: Just noticed the "0 as the filename" special case in dlopen, I'll look into that a bit more tomorrow I guess

MathieuDuponchelle: To be honest after reviewing these new additions, I'm very perplex as to:

* Why the code was complexified to pass this `cmark_mem` structure all around
* Why different nodes in the same tree should have the same "allocator", which is only a thin wrapper calling abort if the allocation failed anyway.

Note that AFAIU on most linux boxes, checking the return of memory allocation syscalls doesn't make much sense anyway.

MathieuDuponchelle: ping @vmg ? I'm more and more skeptical about this ubiquitous cmark_mem being passed around to be honest :(

vmg: I'm not sure I understand the issue here. Every major C library has support for custom memory allocators, and they're implemented just like we did here for CMark. Zlib is just one of thousand examples. http://www.zlib.net/manual.html

This is not a "performance trick", this is a compatibility requirement to use C libraries in real production systems, where the existing platform dictates your allocation behavior. Libraries like CMark need to be wrapped in runtimes and virtual machines that are complex, and most of the time, it's a hard requirement for the library to use the same memory allocator as the rest of the system. As an example, you probably want to use Ruby's memory allocator so the VM can track memory pressure on allocations coming from CMark, or you want to use the CLR's internal allocator to prevent having two split allocation pools inside a .NET application.

I'm well aware it adds complexity to the library, but making portable C code is really hard, whilst making C code that cannot run everywhere is easy but really not useful at all. It's a price we are paying. And I know it's making this PR more complicated for you, but there's isn't much I can do to alleviate the pain. The design we came up with is as clean as we could possibly wire it up.

I'd certainly welcome suggestions on alternate takes for allocator portability, but as @nwellnhof already pointed out, a global allocation setter is a recipe for disaster on large applications that can potentially use the same library from two different contexts. That's why we set the allocator per-parser instead of globally.

MathieuDuponchelle: I'm not really complaining about this pull request, it's now rebased and working, what I do find very peculiar is the introduction of this new requirement in the API, that nodes can only be added to an AST with the same "allocator". This (already pretty old) paper might interest you: 
https://people.cs.umass.edu/~emery/pubs/berger-oopsla2002.pdf

vmg: The same-allocator check for AST nodes is a requirement for a potential memory pool implementation for the nodes. Is it being problematic for this specific PR? If so, feel free to remove that particular check. It can possibly be worked around in the cleanup code for the pool, or I can add it back in the future without much hassle.

MathieuDuponchelle: It's not problematic for my work, as I explained I worked around it, however it *will* be problematic for extensions that do not have access to cmark's private headers, and I would suggest two things if the API is to stay that way:

* Document this new constraint
* Expose API to let extensions query the memory allocator for a given node or parser

jgm: What if, instead of enforcing this constraint in `can_contain`, we simply left it up to the users of the library not to shoot themselves in the foot by using different allocators for parents and children? It seems to me that it would be pretty rare for someone to do this, anyway, and I'm worried that including this check in `can_contain` is making perfectly normal and legitimate uses of the library difficult or impossible.

vmg: Yeah, that's definitely an option as long as we document it. The check is too strict as it stands right now.

jgm: OK, I've removed the check and documented this in 41b8c5b3a4c46d3c2beb1dae6e170c295c19b198

MathieuDuponchelle: Good, good *intently stares at the pull request then at @jgm*

MathieuDuponchelle:
ping again @jgm?

markcarver:
There is now a PHP PECL extension of cmark (https://github.com/krakjoe/cmark), but in order to utilize it within https://github.com/thephpleague/commonmark (which is highly extensible), the underlying `libcmark` library needs to support this.

It would be nice to get some traction on this again.

MathieuDuponchelle:
@markcarver I completely agree that this would be lovely, but we'd need to get @jgm back in here, and validate the design decision made by this PR, which was to implement support for output node types (eg tables) without an official agreed upon input syntax. I am still pretty convinced this is a safe enough decision, but well :)

andrewshadura:
Any updates on this?

MathieuDuponchelle:
ping @jgm

jgm:
Sorry, I just haven't had the bandwidth to devote to this.  This extension mechanism has been implemented in github's fork of cmark.  Those who need extensions can get them there. Since this library purports to be the reference implementation, and it needs to be updated every time the spec is changed, it makes sense to me to keep it simple and not worry about extensions for now at least....

MathieuDuponchelle:
Well let's just close this then

markcarver:
Please don't close this issue. I still believe it's needed as a primary feature in cmark, regardless of how long it may take.

Offering https://github.com/github/cmark-gfm as a "solution" is really much of a solution for https://github.com/krakjoe/cmark or https://github.com/thephpleague/commonmark

Having the ability to extend cmark should be a global ability, not some GFM implementation.

jgm:
I'd also favor keeping this open. I'm not saying the fork was the solution.  I'm just saying that, for now, I don't have the time to deal with it, and keeping the profile of this project slimmer makes it easier for me to keep up with it.

jgm:
I'm open to merging this, really.  But it's a huge patch and surveying it is not simple.  And I'm reluctant to ask @MathieuDuponchelle to rebase it again so it can be reviewed, given my past footdragging on this.  If someone else wants to try to get it in shape, I can take a look, but no guarantees.

markcarver:
I think it would make more sense to split this PR up, i.e. the extension API(s) as a single PR and then the actual subsequent extension implementations into their own individual PRs?

MathieuDuponchelle:
reopened, I had ideas on how to make this more palatable and if there's interest I can give it a go again at some point, I don't think after 3 years there is much of an urgency anyway :)

MathieuDuponchelle:
> I think it would make more sense to split this PR up, i.e. the extension API(s) as a single PR and then the actual subsequent extension implementations into their own individual PRs?

I think the extension API is easier to review alongside extension examples tbh, however I can rework this to disallow mixing / matching multiple extensions from different origins, thus allowing extensions to define their own node types in a "safe" manner, that's basically what github ended up doing iirc

jgm:
Is linking in `dl` going to cause problems for people who are trying to compile the library statically (e.g. on musl) and use it elsewhere?

jgm:
Did GitHub keep the dynamic linking aspect of this, I don't recall?

MathieuDuponchelle:
> Is linking in `dl` going to cause problems for people who are trying to compile the library statically (e.g. on musl) and use it elsewhere?

My plan would get rid of the dl linking entirely

MathieuDuponchelle:
> Did GitHub keep the dynamic linking aspect of this, I don't recall?

I think they didn't, but that was a long time ago :) The problem without dynamic linking is that you end up with two solutions for people who want extensions:

* Maintain their own cmark extension, compile in your extensions (meh)

* Maintain extensions in cmark itself (more maintaining burden on the cmark side, but you're at least guaranteed you won't break extensions and it makes users life easier)

jgm:
Can you clarify this?  You want to avoid linking in dl, but still have dynamic plugins?  How is it done?  (I'm sorry, I've forgotten a lot of the context.)

MathieuDuponchelle:
> Can you clarify this? You want to avoid linking in dl, but still have dynamic plugins? How is it done? (I'm sorry, I've forgotten a lot of the context.)

No, those two potential solutions were in the event that cmark wouldn't allow for third-party / dynamic plugins, hence not use dl / equivalent for other platforms.

I'm on the fence as to whether I want dynamic extensions or not, as in my specific case I have extensions for which it would make no sense to maintain them in cmark itself (eg, inline links to symbol names with `#foo` or `foo()` for function names, this is syntax from gtk-doc which I need to support). This means that without dynamic plugins, my only option will be to keep maintaining a fork, although with limited chances for conflicts when rebasing, which is already an improvement compared to my current situation :)

jgm:
> inline links to symbol names with #foo or foo() for function names

I'm pretty sure you don't need anything as heavyweight as an extension to support these.  These could be handled with a transformation of the AST after parsing, no?

MathieuDuponchelle:
> I'm pretty sure you don't need anything as heavyweight as an extension to support these. These could be handled with a transformation of the AST after parsing, no?

nope, not if I care about having proper parsing at least :) I don't remember the exact details, but a syntax extension is the correct approach here. Besides, these are just examples really, and syntax extensions are not heavyweight, they're actually more performant than whatever a posteriori, fragile AST parsing I could come up with.

Tregman3:
> Reference discussion: #100
> 
> I just rebased the branch on master and tested it, make passes with no new warnings all along the branch, valgrinding a table test shows no leaks.
> 
> I applied the initial review comments from @jgm and @nwellnhof , which were to not expose cmark_strbuf, which is thus included in the sources of the reference "core extension", and to not expose API using cmark's bool type, which is replaced by integers, rendering installation of config.h unnecessary.
> 
> The design implemented is the second one I proposed in the above discussion, here's a brief summary:
> 
> * A new structure is exposed, `cmark_syntax_extension`, which exposes hooks that match the block and inline parsing strategies outlined in http://spec.commonmark.org/0.25/#appendix-a-parsing-strategy .
> * This proposal does not cover the implementation of new block types, instead syntax extensions need to create blocks with types defined by cmark itself, including `CMARK_NODE_CUSTOM_BLOCK` and `CMARK_NODE_CUSTOM_INLINE`, this already covers a wide range of use cases.
>   That is because in my opinion, we want cmark to guarantee that the output AST will always be correctly structured (with the exception of custom inlines and blocks), with no "impossible" or "maningless" block nesting situations.
>   An interesting enhancement would be to refactor the custom inline and block types to let extensions provide virtual methods such as "can_contain" for them, but this will be somehow fragile as "inter-extension negotiation" will not be possible.
> * An important part of this proposal is that cmark implement and expose block types for which it does not yet have a formal syntax specification, such as tables or strikethroughs. I believe this is desirable because this will allow multiple syntax extensions to compete for a given desired output, without tying cmark's specification to a specific syntax, while still making sure the produced output is correct. It makes sense when looking at it as a testing ground if you will.
> 
> It is important to state that this design doesn't prevent us from implementing negotiation in the future if we so desire.
> 
> I think this branch is now approaching what I'd consider mergeable quality, with the exception of the plugin code, which I think is mostly correct on Linux, but will not work on Windows. If @nwellnhof is interested in tackling the abstraction for Windows, then all the best, otherwise the code can be merged without plugin loading / discovery for now, as it is already useful for people using cmark as a library (such as I).
> 
> Last thing, this last rebase round has been pretty involved, and I'd very much like (if we all agree that extensions are desirable) that merging this branch be made a priority, I will of course address reviews in a timely manner.

N

parsamparham (approved):

Lacters (approved):

turbodt:
I've been reading both this thread and [#100](https://github.com/commonmark/cmark/issues/100).
I'm not sure what has happened during these 8 years.
I'm not an experienced contributor.
I do not know much about collaboratively building software.

Last week I was looking for a C library to parse MarkDown into an AST, and that library was perfect.
Now I would like to implement a custom flavour for my use case.
And (if I understood correctly) this library cannot handle this necessity and the changes that enable that feature are in a forked repository which is +350 commits behind master.

I guess it is a very lucky coincidence that last approval for this PR was less than a month ago.

What I want to communicate with this message is:

> Despite a lot of time has passed, this topic is still useful to somebody.

With my current experience in C development I don't think I cannot contribute to it - but I'll try my best.

Thanks for reading.

jgm:
> I guess it is a very lucky coincidence that last approval for this PR was less than a month ago.

That doesn't mean much, because the people who approved it are not associated with this project.

This project aimed to be a reference implementation for the commonmark spec, and to be easily embeddable in other projects (no dependencies, can be statically linked).  Adding an extension framework is in tension with the first goal (though perhaps compatible), and if the framework uses dynamic linking (as this PR does), with the second.

If you want to think about how an extension mechanism should work, you're welcome to, and I don't rule out including one, but don't get hopes up!

PR #127, opened 2016-05-23
buffer: proper safety checks for unbounded memory

vmg:
Hey @jgm! Long time no chat. :)

I've been doing some security review on the library. I have some concerns about the way we're handling buffer overflows, so here's a proposed commit. Message as follows:

---

The previous work for unbounded memory usage and overflows on the buffer
API had several shortcomings:

1. The total size of the buffer was limited by arbitrarily small
precision on the storage type for buffer indexes (typedef'd as
`bufsize_t`). This is not a good design pattern in secure applications,
particualarly since it requires the addition of helper functions to cast
to/from the native `size` types and the custom type for the buffer, and
check for overflows.

2. The library was calling `abort` on overflow and memory allocation
failures. This is not a good practice for production libraries, since it
turns a potential RCE into a trivial, guaranteed DoS to the whole
application that is linked against the library. It defeats the whole
point of performing overflow or allocation checks when the checks will
crash the library and the enclosing program anyway.

3. The default size limits for buffers were essentially unbounded
(capped to the precision of the storage type) and could lead to DoS
attacks by simple memory exhaustion (particularly critical in 32-bit
platforms). This is not a good practice for a library that handles
arbitrary user input.

Hence, this patchset provides slight (but in my opinion critical)
improvements on this area, copying some of the patterns we've used in
the past for high throughput, security sensitive Markdown parsers:

1. The storage type for buffer sizes is now platform native (`ssize_t`).
Ideally, this would be a `size_t`, but several parts of the code expect
buffer indexes to be possibly negative. Either way, switching to a
`size` type is an strict improvement, particularly in 64-bit platforms.
All the helpers that assured that values cannot escape the `size` range
have been removed, since they are superfluous.

2. The overflow checks have been removed. Instead, the maximum size for
a buffer has been set to a safe value for production usage (32mb) that
can be proven not to overflow in practice. Users that need to parse
particularly large Markdown documents can increase this value. A static,
compile-time check has been added to ensure that the maximum buffer size
cannot overflow on any growth operations.

3. The library no longer aborts on buffer overflow.  The CMark library
now follows the convention of other Markdown implementations (such as
Hoedown and Sundown) and silently handles buffer overflows and
allocation failures by dropping data from the buffer. The result is
that pathological Markdown documents that try to exploit the library
will instead generate truncated (but valid, and safe) outputs.

All tests after these small refactorings have been verified to pass.

---

NOTE: Regarding 32 bit overflows, generating test cases that crash the
library is trivial (any input document larger than 2gb will crash
CMark), but most Python implementations have issues with large strings
to begin with, so a test case cannot be added to the pathological tests
suite, since it's written in Python.


jgm:
These all sound like good changes to me, but I'd like to get feedback from @nwellnhof too.

kainjow:
> The result is that pathological Markdown documents that try to exploit the library will instead generate truncated (but valid, and safe) outputs.

Can tests be added for this, to prove it's working as expected? Could write a script that generates a 32+MB file, instead of committing a large file.

vmg:
@kainjow: Done! I've added a memory exhaustion test that just pipes content into a temporary file. 👌

nwellnhof:
I agree that cmark should impose some limits for untrusted input to avoid memory exhaustion. But the strbuf code seems like the wrong place to me. The cmark parser uses a separate strbuf for every block node, so you could simply create a document with many paragraphs to exhaust memory. If you're serious about this, you also have to consider memory used for cmark_nodes. The worst case I can come up with is a document like

    a

    a

    a

    ...

which creates two nodes (a paragraph and a text node) for every three input bytes. Since a cmark_node uses about 100 bytes, this results in about 150 bytes of memory allocated for every input byte. So a robust solution should limit the size of the input at the parser input level and the total number of nodes created. A maximum of 100,000 - 200,000 nodes seems reasonable. It would also be nice if these limits can be adjusted at runtime. (We probably need a limit for the reference map, too.)

With such a limit in place, the strbuf code only aborts if there's a real out-of-memory situation. In this case, aborting isn't too bad. On a typical Linux system, the OOM killer would start to abort processes anyway.

A better solution is of course to return error codes, but these errors should be passed up all the way to the caller of a public cmark function. If we ignore error conditions internally, the library has undefined behavior in case of OOM errors. It won't crash but it can truncate input and output buffers in an unpredictable way. IMO, this is worse than aborting.

I don't have a strong opionion on whether bufsize_t should be typedef'd to ssize_t or int. But we should consider using a 32-bit type on 64-bit systems internally to save memory.

vmg:
> With such a limit in place, the strbuf code only aborts if there's a real out-of-memory situation. In this case, aborting isn't too bad. On a typical Linux system, the OOM killer would start to abort processes anyway.

The `strbuf` code should never `abort`. Please stop considering that option -- it makes CMark impossible to use in real production systems which are tuned for security (e.g. running with custom memory allocators, with a custom ulimit, etc). A library with `abort` is a terrible software citizen. 😞

We can handle OOM conditions gracefully like this PR already does -- the error condition is being handled transparently, not ignored, and there's no undefined behavior. This is how we have written the existing parser on GitHub.com and Reddit.com and it covers the vast majority of DoS attacks that we've seen live. Bubbling up error codes is an option, but generally speaking, we don't want to 500 or "crash" at rendering any documents when we could be silently truncating them. The `strbuf` code is the ideal place to have these limits, particularly in a library that can output to all kinds of different renderers and formats. Buffer code is the main source of allocations in a running process and the only path that the renderers have in common.

I agree that a limit on nodes would also be a good thing to have, but that's a bit of a major undertaking. I can work on that on another PR.

> But we should consider using a 32-bit type on 64-bit systems internally to save memory.

8 bytes saved per buffer are hardly a compelling case, particularly when the cost is having a "safe strlen" and a bunch of different helpers. Capping the buffer size at 31 bits brings more trouble than it saves us. 😓

nwellnhof:
> A library with abort is a terrible software citizen.

I agree but I don't think that ignoring errors silently is the solution.

> We can handle OOM conditions gracefully like this PR already does -- the error condition is being handled transparently, not ignored, and there's no undefined behavior.

It isn't undefined behavior in the sense of the C language specification, but the error is definitely ignored. In this case, it is almost impossible to specify how your API behaves. So an honest API documentation would look like:

> `cmark_markdown_to_html(...)`
>
> Returns the input Markdown rendered as HTML most of the time. If an out-of-memory error occurs, this function returns garbage and we don't tell you if this happened. All we can guarantee is that the function won't crash.

There are use cases where this is acceptable, but I prefer a library that reports errors in one way or another. During development, an OOM error is often caused by bugs further up the callstack which are hard to debug if you don't get any error messages. Ignoring errors can also lead to other security bugs. For example, if you rely on `strbuf_puts` to terminate a buffer with certain characters, and the function silently ignores your request. This isn't likely in the context of cmark, though.

The only real solution is to bubble up error conditions everywhere in the library.

> 8 bytes saved per buffer are hardly a compelling case, particularly when the cost is having a "safe strlen" and a bunch of different helpers. Capping the buffer size at 31 bits brings more trouble than it saves us.

If you want to limit the size of strbufs anyway, it shouldn't make a difference whether you cap the buffer size at INT_MAX or anywhere else.

vmg:
Alright, let's try this instead: since you're really set on `abort`, I've added an API to set a custom OOM handler. The library now has the same default behavior as before (abort on OOM, abort on buffer overflow), but you can now choose to override that when integrating it on your own applications.

While I was at it, I added proper OOM handling to all other heap memory allocations, where I noticed we weren't really checking for `NULL` pointer returns from `malloc`, `calloc` and so on. All memory allocations will trigger the OOM handler when running out of memory.

I've also removed the bad OOM handling patterns that we had in several files. There was this common idiom where we'd check for a `NULL` pointer right after calling `calloc` to e.g. initialize the contents of the pointer, but right afterwards we'd return the potentially `NULL` pointer anyway. That obviously doesn't prevent any crashes, it only bubbles the crash one level up in the callstack, where the caller assumes a non-`NULL` pointer.

I think this is a good compromise that simplifies the library quite a bit and fixes the immediate issue for people who cannot run a library that `abort`s (they just need to set a custom handler now). Hence, this PR takes care of OOM and overflow situations, and leaves the "maximum output size" and "maximum number of allocated nodes" problem for another day.

nwellnhof:
> since you're really set on abort

Like I said, I'm not set on abort. If I find some time, I can prepare a pull request that adds proper error handling to the library.

vmg:
>  I can prepare a pull request that adds proper error handling to the library.

Sounds like a very major undertaking, since the only error cases will be OOM errors, and the library wasn't really designed with that in mind. I would definitely love to see that project come to term, though. 😄 

@jgm: Nonetheless, can we merge this PR meanwhile? It fixes the immediate issue in a pretty clean way.

vmg:
...Alternatively, it occurs to me that we could also implement support for a "custom memory allocator", so people can use the memory allocator of their runtime and add their own checks for OOM situation...

I think that's an interesting compromise. Although I would love to see "proper error handling", I'm afraid that refactoring will end up being _massive_, and the only error cases that are going to be handled are when `malloc` or `realloc` return `NULL`. Adding support for custom memory allocators would let us keep the critical paths of the library as-is (without more checks and without having to bubble up allocation failures) and users could use the memory allocator for their runtime so e.g. you can pass Ruby's memory allocator that will raise an exception when it runs out of heap memory.

@nwellnhof What do you think about this idea?

nwellnhof:
@vmg I think it's a good idea. But the callbacks should be stored in a per-parser context, not in a global variable. Mutable globals can cause huge problems if a library has multiple users in a single process. Consider the following scenario which has plagued libxml2 for years: You have an Apache httpd process with mod_php, mod_ruby, and mod_perl. Each of these modules could load libcmark for the respective language bindings. If the language bindings try to set different global callbacks, all hell breaks loose. I think we need something like

    typedef void*
    (*cmark_malloc_fn)(size_t size);
    typedef void
    (*cmark_free_fn)(void *ptr);

    cmark_parser*
    cmark_parser_new_with_allocator(int options,
                                    cmark_malloc_fn *custom_malloc,
                                    cmark_free_fn *custom_free);

    struct cmark_parser {
        ...
        cmark_malloc_fn malloc;
        cmark_free_fn free;
        ...
    };

So you'd have to use the `cmark_parser` interface to use custom allocators. Unfortunately, this means that cmark has to pass the `cmark_parser` to every internal function that allocates memory.

jgm:
+++ Vicent Marti [May 25 16 04:31 ]:
>   [1]@jgm: Nonetheless, can we merge this PR meanwhile? It fixes the
>   immediate issue in a pretty clean way.

It might have to wait til next week, as I'll be going out of
town, and it seems as if there's still a productive
discussion about which way to go.

vmg:
> But the callbacks should be stored in a per-parser context

Yeah that's reasonable. Also a good thing since it means we can possibly pool the allocations for nodes. I'm gonna get started on implementing this. Let's see how hard it turns out to be.

nwellnhof:
:+1: Allocating nodes from a memory pool is a great idea.

MathieuDuponchelle:
For what it's worth, the GLib has a good (and well-tested) implementation of a memory pool, see <https://developer.gnome.org/glib/stable/glib-Memory-Slices.html>, just saying ;)

vmg:
Check out https://github.com/jgm/cmark/pull/133

jgm:
@vmg I'm uncertain whether #133 really achieves all the objectives you had in this original PR.
Could you comment?

For example, in `buffer.c` lines 44-45 we currently have
```
  if (target_size > (bufsize_t)(INT32_MAX / 2))
    abort();
```
where your original PR (#127) had
```
    return false;
```

I don't see how this abort risk is removed by the custom memory allocators you added in #133.

PR #132, opened 2016-06-01
Changed type from int to size_t to fix implicit type conversion warning

BenedictC:
I encountered this warning when building for x64_86 with Xcode 7.3.


jgm:
Thanks

PR #133, opened 2016-06-02
Implement support for custom memory allocators

vmg:
Supersedes https://github.com/jgm/cmark/pull/127

As discussed on the previous PR, here's a proposal on a supporting custom memory allocators. As you can see I've wired up the `cmark_mem` structure throughout the parser, I believe without increasing complexity or memory usage needlessly. I haven't implemented yet pooling allocators but it should be trivial now that we're passing a "memory" structure everywhere where we allocate memory.

I'd love feedback on this design. I'm moderately happy with it. The external API hasn't been broken, and the "default APIs" will continue working as before, using a default allocator (system `malloc` + `abort` on OOM). Every single node is now also aware of its allocator, which means that we can choose an allocator when creating new nodes and check that you cannot insert a node from a specific allocator into a document tree created by another allocator.

To offset the slight memory increase in the `node` structure, I've put some light memory diet -- although we could save quite a few bytes if we did something smarter with the `prev`, `next`, etc links in each structure.

cc @jgm @nwellnhof


jgm:
Sorry about the test failures. I discovered yesterday that the roundtrip test was defective -- the test suite wasn't catching errors.  When I fixed that, we got some errors.  I've now revised the roundtrip test and fixed a small bug in the commonmark renderer, so all tests should pass again.

Can you rebase against master and we'll try the CI tests again?

vmg:
@jgm: No worries! Rebased.

src/cmark.h
```diff
@@ -4,6 +4,7 @@
 #include <stdio.h>
 #include <cmark_export.h>
 #include <cmark_version.h>
+#include "memory.h"
```
nwellnhof: This line should be removed since `memory.h` is gone. It didn't break because there is a system header `memory.h`. Or did you mean to include the system header?

nwellnhof:
Looks good. Maybe names like `*_new2` could be made more descriptive.

Regarding the Windows breakage: The MSVC 12 build can probably be fixed by including `stdint.h` in `buffer.h`. The MSVC 10 build breaks because we compile in C++ mode with older MSVC versions and C++ is more strict regarding enums. You need an explicit cast to `cmark_node_type` in `cmark_node_get_type`.

src/cmark.h
vmg: No, this is a leftover. Thanks for pointing it out. :)

vmg:
Thanks for the feedback! I'm going to fix the MSVC issues (it may be a bit laborious because I don't have a Windows machine here, but I'll just keep using CI).

> Looks good. Maybe names like *_new2 could be made more descriptive.

Yes, I tried other naming conventions, but they were pretty bad. `new_with_mem` sounds good to you?

nwellnhof:
`new_with_mem` sounds good. If you don't have a Windows machine, you can leave the MSVC issues to me.

jgm:
I haven't had a chance to review this in great detail, but the basic idea sounds fine to me, and I trust the two of you much more than myself on this particular issue.  So if you're in agreement that this is a good change, I'll be happy to merge it.

Perhaps it makes sense to wait for the Windows fixes and possible renaming.  Do we really want `cmark_node_new2` and `cmark_parser_new2`?  I'd prefer something like `cmark_node_with_allocator`.

Also, this change exposes a struct in the public API (`cmark_mem`).  We'd avoided that so far, instead exposing getter and setter functions that operate with standard C types and enums.  This makes the API easier to bind to from dynamic languages, and gives us freedom to modify the underlying data structures without changing the API.  It's worth thinking about whether we want to expose the `cmark_mem` structure, as opposed to just having two parameters to `cmark_node_new2` and `cmark_parser_new2`, one for calloc and one for free.

I ran the benchmark and got about a 10% performance drop from this change.  I suppose that's not unexpected.  Could we conceivably gain that back again by using a specially tuned custom allocator?  I confess I don't have a very good sense for what sorts of custom allocators could be used.  If you have an example, that might be helpful.

nwellnhof:
>  It's worth thinking about whether we want to expose the cmark_mem structure, as opposed to just having two parameters to cmark_node_new2 and cmark_parser_new2, one for calloc and one for free.

This would work if the `cmark_mem` struct was only needed by the parser. But it really has to be part of `cmark_node`, so I don't see an easy way around allocating this struct outside of `cmark`. This is also a feature that can't be exposed to dynamic languages anyway.

> I ran the benchmark and got about a 10% performance drop from this change.  I suppose that's not unexpected.

No, I really didn't expected that. I got an even bigger drop of about 15% on 32-bit Linux. It turns out that (on my machine) this drop comes only from changing `realloc` in `buffer.c` to a `calloc`, `memcpy`, `free` sequence. I hacked `realloc` back in, and performance was back to the previous level. So it seems that the custom allocator needs a `realloc` function.

vmg:
Indeed, removing `realloc` as to simplify the allocation API was definitely a mistake. I'm going to bring that back in, that performance regression is terrible. Sorry about that!

I'm also going to fix the MSVC errors this morning. Thanks for your patience. 👌

vmg:
Alright, I got the Windows builds working and I've tackled all feedback. Somehow the roundtrip test on MSVC12 is still failing tho... What gives?

nwellnhof:
The test failure on MSVC12 is an unrelated issue fixed in #134.

jgm:
No idea why it fails on MSVC12 and passes on MSVC10.

I don't have an MSVC12 environment to test on.  @nwellnhof
can you see what is going on here?

+++ Vicent Martí [Jun 06 16 06:44 ]:

jgm:
Great, this has been merged, so @vmg if you rebase again,
it should pass.

+++ Nick Wellnhofer [Jun 06 16 09:33 ]:

vmg:
It's green! :)

jgm:
Thanks, I've merged all of these commits into master.

I'd still be interested in an example of a custom memory allocator; this would be helpful for documentation.

PR #134, opened 2016-06-06
Fix character type detection in commonmark.c

nwellnhof:
- Implement cmark_isalpha.
- Check for ASCII character before implicit cast to char.
- Use internal ctype functions in commonmark.c.

Fixes test failures on Windows and undefined behavior.


jgm:
Great, thanks.

PR #137, opened 2016-06-27
CMake fixes

foonathan:
Just two fixes to use the library better as CMake subproject (`add_subdirectory()`).

1. Export the targets for installation. This will create a file generating an IMPORT target, required for e.g. `find_package()` and allows linking the cmark targets to your own targets if your own targets are installed.

2. Remove a developer warning by suppressing the policy (see comments).


CMakeLists.txt
```diff
@@ -29,3 +29,6 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Release" CACHE STRING
   "Choose the type of build, options are: Debug Profile Release Asan Ubsan." FORCE)
 endif(NOT CMAKE_BUILD_TYPE)
+
+# export targets
+install(EXPORT cmark DESTINATION lib/cmake)
```
jgm: Is this really supposed to be `lib/cmake`?

jgm: Also note the comment at line 49 of src/CMakeLists.txt.

foonathan:
CMake has a couple of places where it searches for config files generated by the install(EXPORT) command. <prefix>/lib/cmake is one of them, useful if your project doesn't add its own subfolder for the libraries (like this here does).

I can change that to ${LIB_INSTALL_DIR}/cmake, didn't see the comment.

foonathan:
When is this PR going to be merged?

jgm:
I'm just worried about assuming that the build machine will always have a `$LIB_INSTALL_DIR/cmake` directory.

Is that universally true, even on Windows e.g.?

foonathan:
The common CMake convention is that you install and export libraries under `lib/project`. But to suppport installing to `lib` directly you can also export under `lib/cmake`.

And the [documentation for find_package()](https://cmake.org/cmake/help/v3.6/command/find_package.html) lists `<prefix>/lib/cmake`. It doesn't assume that the project has such a directory, it will be created when needed (afaik).

jgm:
OK, I've merged now!

foonathan:
Thanks! If you run into any CMake issues related to that, I'll be happy to help.

PR #138, opened 2016-06-30
cmark_node_get_literal support for NODE_IMAGE

k06a:


jgm:
This just returns the URL.  If you want the URL, there is a separate API function for that already.

PR #143, opened 2016-07-13
link fix

tst2005:


jgm:
Thanks - wouldn't be needed if GitHub used CommonMark!

PR #153, opened 2016-09-11
autolink scheme can contain digits

gaborcsardi:
According to the specs: http://spec.commonmark.org/0.26/#scheme

> For purposes of this spec, a scheme is any sequence of 2–32 characters 
> beginning with an ASCII letter and followed by any combination of ASCII
> letters, digits, or the symbols plus (”+”), period (”.”), or hyphen (”-”).

### DIFF
--- a/src/scanners.re
+++ b/src/scanners.re
@@ -71,7 +71,7 @@ bufsize_t _scan_at(bufsize_t (*scanner)(const unsigned char *), cmark_chunk *c, 
   in_single_quotes = ['] (escaped_char|[^'\x00])* ['];
   in_parens        = [(] (escaped_char|[^)\x00])* [)];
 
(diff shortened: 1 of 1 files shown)


nwellnhof:
Also, allowing a space at the beginning of a scheme seems wrong.

jgm:
I'll change the space and also regenerate scanners.c.

gaborcsardi:
@jgm Thanks much!

jeroenooms:
@gaborcsardi do you need a patch release for the [commonmark](https://cran.r-project.org/web/packages/commonmark/index.html) R bindings or can this wait until the next stable release of the `cmark` library ?

gaborcsardi:
@jeroenooms Thanks, not urgent, I think. 

The current plan is use sg else for links, not the autolink feature, because the schema does not allow dots and underscores, so we cannot use it as a package name.

PR #154, opened 2016-09-19
Extensions redux

kivikakk:
Hi there!

I've taken the work in #123 and rejigged it a bit. At GitHub we're currently using a [Sundown](https://github.com/vmg/sundown)-based parser/renderer, but it's not super extensible. So, we've decided to roll out CommonMark to replace it.

Here are some of the changes to #123 in this PR:
- Took out the shared object searcher. I don't think library code should searching and loading objects dynamically at runtime. Instead, you as the user register whatever plugins you might have linked in yourself. (Maybe _you_ loaded it dynamically — that's not something for a Markdown library to do, imo.)
- Expanded the extension interface enough such that the two existing plugins (table, strikethrough) can be implemented without any changes to the core code. #123 had table specific code in the core (outside of the shared object), which limited its usefulness. This branch means no change to core code for implementing said. Extensions can register their own node types, their own renderer functions for said node types, etc.
- Fixes the Windows build.
- Adds tests.
- Adds an autolink extension.
- Adds a whitelist extension for the HTML renderer.

This functionality has all been exposed as opt-in in the Ruby gem [commonmarker](https://github.com/gjtorikian/commonmarker), which is our primary interface.

### DIFF
--- a/extensions/autolink.c
+++ b/extensions/autolink.c
@@ -27,7 +27,7 @@ static int sd_autolink_issafe(const uint8_t *link, size_t link_len) {
 }
 
 static size_t
-autolink_delim(uint8_t *data, size_t link_end, size_t max_rewind, size_t size)
+autolink_delim(uint8_t *data, size_t link_end)
 {
 	uint8_t cclose, copen;
 	size_t i;
@@ -39,7 +39,7 @@ autolink_delim(uint8_t *data, size_t link_end, size_t max_rewind, size_t size)
 		}
 
 	while (link_end > 0) {
-		if (strchr("?!.,:", data[link_end - 1]) != NULL)
+		if (strchr("?!.,:*_", data[link_end - 1]) != NULL)
 			link_end--;
 
 		else if (data[link_end - 1] == ';') {
@@ -106,7 +106,7 @@ autolink_delim(uint8_t *data, size_t link_end, size_t max_rewind, size_t size)
 
 		if (closing == opening)
 			break;
-		
+
 		link_end--;
 	}
 
@@ -165,7 +165,7 @@ static cmark_node *www_match(cmark_parser *parser, cmark_node *parent, cmark_inl
   while (link_end < size && !cmark_isspace(data[link_end]))
     link_end++;
 
-  link_end = autolink_delim(data, link_end, max_rewind, size);
+  link_end = autolink_delim(data, link_end);
 
   if (link_end == 0)
 		return NULL;
@@ -232,7 +232,7 @@ static cmark_node *email_match(cmark_parser *parser, cmark_node *parent, cmark_i
       (!cmark_isalpha(data[link_end - 1]) && data[link_end - 1] != '.'))
 		return 0;
 
-	link_end = autolink_delim(data, link_end, max_rewind, size);
+	link_end = autolink_delim(data, link_end);
 
   if (link_end == 0)
 		return NULL;
@@ -283,7 +283,7 @@ static cmark_node *url_match(cmark_parser *parser, cmark_node *parent, cmark_inl
 	while (link_end < size && !cmark_isspace(data[link_end]))
 		link_end++;
 
-	link_end = autolink_delim(data, link_end, max_rewind, size);
+	link_end = autolink_delim(data, link_end);
 
 	if (link_end == 0)
 		return NULL;
@@ -304,7 +304,8 @@ static cmark_node *url_match(cmark_parser *parser, cmark_node *parent, cmark_inl
 }
 
 static cmark_node *match(cmark_syntax_extension *ext, cmark_parser *parser, cmark_node *parent, unsigned char c, cmark_inline_parser *inline_parser) {
-  if (cmark_inline_parser_in_bracket(inline_parser, false))
+  if (cmark_inline_parser_in_bracket(inline_parser, false) ||
+      cmark_inline_parser_in_bracket(inline_parser, true))
     return NULL;
 
   if (c == ':')
--- a/test/extensions.txt
+++ b/test/extensions.txt
@@ -165,9 +165,18 @@ scyther@pokemon.com
 
 www.github.com www.github.com/á
 
+www.google.com/a_b
+
+**Autolink and http://inlines**
+
+![http://inline.com/image](http://inline.com/image)
+
 .
 <p>: <a href="http://google.com">http://google.com</a> <a href="https://google.com">https://google.com</a></p>
 <p><a href="http://google.com/%C3%A5">http://google.com/å</a> <a href="http://google.com/%C3%A5">http://google.com/å</a></p>
 <p><a href="mailto:scyther@pokemon.com">scyther@pokemon.com</a></p>
 <p><a href="http://www.github.com">www.github.com</a> <a href="http://www.github.com/%C3%A1">www.github.com/á</a></p>
+<p><a href="http://www.google.com/a_b">www.google.com/a_b</a></p>
+<p><strong>Autolink and <a href="http://inlines">http://inlines</a></strong></p>
+<p><img src="http://inline.com/image" alt="http://inline.com/image" /></p>
 ````````````````````````````````


jgm:
The object-loading code and the extension-specific changes in core were two things I didn't like about #123, so this seems a step in the right direction.  However, I'd be interested in comments from @MathieuDuponchelle and @nwellnhof.

This is a *big* code revision.  It's hard to see the forest for the trees.  To ease review, could you please rebase it so that (a) all of the changes are on top of current master, (b) commits are reordered and consolidated into (a smaller number of) logically coherent commits?  Part of the difficulty with #123 was that it was hard to survey.

Also note that travis builds are failing due to a warning.

kivikakk:
Travis was failing due to an older Valgrind on the Linux instances. That's been fixed up.

I'll look into cleaning up the PR. It is rather large; I probably won't touch commits from #123

nwellnhof:
We should really try to split this up into smaller chunks. First of all, I'd suggest to tackle block and inline extensions separately. Then, *before writing any new code,* we should start by discussing the internals.

## Discuss changes to core data structures

### How to identify extension nodes

The new PR improves upon the previous approach by generating numeric ids for new node types dynamically. Personally, I'd still prefer to have two single node types `CMARK_NODE_EXTENSION_BLOCK` and `CMARK_NODE_EXTENSION_INLINE` for all extension nodes and to identify nodes by name (for example using URIs like in XML namespaces). This makes comparisons a little slower but avoids a couple of problems with generating ids dynamically (thread safety, registering the same extension for multiple users, identifying nodes without having access to the generated id). 

### Render hooks and extension-specific node data

I'd propose to add two new fields to the `cmark_node` struct for extension nodes. An opaque `void *` like in the new PR for extension-specific data and a pointer to a struct with the following elements:

    struct cmark_ext_node_type {
        // A unique name for the node type, possibly a URI to avoid clashes.
        char *name;
        // Hooks for the HTML renderer.
        void (*html_on_enter)(cmark_node *node, ...);
        void (*html_on_exit)(cmark_node *node, ...);
        // Destructor.
        void (*destroy)(cmark_node *node);
    };

The PR adds a separate field for the destructor and stores other callbacks in the extension struct. I think it's cleaner to have separate struct for each node type which can contain callbacks and other static data .

## Discuss public extension API

I think the best approach is to look at the core node types and restructure the code in `blocks.c` and `inlines.c` as if core nodes were extension nodes themselves. I recently did this for block nodes and realized that we only need a single hook in `open_new_blocks`. If a new (extension or core) block is found, a new node can be created and three callbacks are needed for further processing:

* **check_open** which checks whether the block is still open. This callback is invoked from `check_open_blocks` and replaces the switch statement with a call to a function pointer.
* **add_line** which adds the end of the current line to the block's content. This callback is invoked from `add_text_to_container` and replaces the switch statement with a call to a function pointer.
* **finalize** which finalizes a block. This callback is invoked from `finalize` and replaces the switch statement with a call to a function pointer.

These callbacks only need access to the current line and offset in `cmark_parser`. They don't have to be  registered globally and can be passed to the parser for each new block.

This isn't too different from the PR, but it seems a bit simpler to me. This also makes it straightforward to support core blocks inside extensions blocks. I'm not sure how this would be handled in the PR. I can publish my changes to the block parser for review if anyone is interested.

I haven't looked at the inline extension API yet. It will probably be a lot more complicated than the API for extension blocks. I only noted that the PR exposes the delimiter struct.

## Agree on the design

As I already mentioned in #123, I'm still not happy with some design decisions. See "Discuss changes to core data structures" above where I reiterate my criticism.

## Implement

This should really be the last step.

vmg:
These are all very good suggestions, @nwellnhof! Unfortunately we're short on time as we're rolling out these changes to the website in the upcoming weeks, so we can't afford to sit down and discuss the extensions API for the hundredth time. Making progress on this stuff has been _extremely painful_ so far.

The truth is that CMark is not gaining traction because nobody is using it on production, and nobody is using it on production because it doesn't support any of the features that the large websites (like us at GitHub) need to use it.

So, to speed up things, we'll be publishing soon our own fork of CMark. Hopefully you can pick up from there all the bits and pieces until this CMark implementation becomes more usable, or maybe other websites will just start using our fork -- either way we all win because it means the CMark standard is getting adoption. Thank you and everybody involved for the fantastic work so far.

MathieuDuponchelle:
woot :) I'll take some time to look at it later today, thanks a bunch :)

jgm:
I have to admit, adding an extension mechanism hasn't been a big priority for me.  The primary aim of this library is to be a reference implementation for the CommonMark spec and a demonstration of algorithms for efficiently parsing the syntax defined by the spec.  It was never intended to be the only implementation.  (See Markdig and markdown-it for extensible CommonMark implementations in C# and JavaScript.)  In principle I like the idea of an extension mechanism, but everything I've seen so far has been too complex for my taste, and I haven't had time to try to come up with anything I like better.  (Maybe it's not possible.)  Publishing a fork that explores ideas for extensibility is a reasonable thing to do.

That said, I completely agree with the things @nwellnhof has said here, and I'm a bit taken aback by @vmg's response.  As a user of GitHub who has often been plagued by Markdown-related problems, I think that GitHub _ought_ to be interested in thinking through the details of the extension architecture and API very carefully before rushing code into production. It seems foolish to give up the chance to discuss this with people who have thought quite a lot about it.

MathieuDuponchelle:
OK, I'll take the time to formulate a more constructive answer tomorrow, but I'm not so enthusiastic anymore, @vmg saying you're gonna use your own fork is fine and a good thing to do, and I'm happy if my code can be useful in some manner to you folks, but seriously this pull request is a joke, and stating 13 hours after you made it "maybe other websites will start using our fork" is just ruthlessly bad-mannered in my very humble opinion, especially coming from someone representing a company that does have the power to carry out this kind of aggressive fork, which is what that would be, no matter how you may sugar-coat it.

More constructively, the reason why I didn't complain about the perceived lack of reactivity from upstream is that the reduction of scope needed to be done, I would have indeed liked a bit more guidance from @jgm, but we all have our time constraints and I respect that faced with 20+ large commits it's hard to take the time to sit down and go through them.

However the code dump that's proposed there is the exact opposite of "redux", with partial manual reverts, obviously WIP patches such as "Super hack to move table state out of HTML", I don't know how you guys even have the nerve to propose this seriously!

**Important** the main reason why @jgm was unhappy with my request was a very meaningful design decision that I made, which was to not allow registration of new types of nodes from the outside, because then mixing and matching extensions from different providers that wouldn't know about each other, and thus couldn't specify *correctly* containing rules, was not possible.

That use case might have been a bit too extreme, and imposing severe restrictions on the design, as a side note that's for this use case that I half-implemented the plugin loading mechanism.

As I don't even need this for my use case, as I want to expose a specific set of extensions, over which I have complete control, much like I suppose you guys at github do, I decided some time ago I'd undertake a rework, and provide an *actual* redux version which upstream would be happy with.

You guys should have come discuss on #123 to figure out how to approach this task which you visibly agree needed doing, I'd have been more than happy to help and we would have ended up with something more mergeable than at the beginning, in its state that branch is, in my opinion, way less mergeable than the original one, especially as it entirely contains it then partly deconstructs it ...

vmg:
@jgm @MathieuDuponchelle: I thoroughly apologize, I did not mean to offend or be rude with my comments. I'm just stating the reality of our schedule/timeframes which is sadly not up to me.

I wish we could all agree on the design a beautiful extensions API, but it seems very unlikely that we'll do so on a reasonable timeframe -- not because a lack of will or effort, but as @jgm points out, because this is a particularly hard problem.

I'm still thoroughly convinced that the best thing we can do for CMark is getting it running on one of the largest websites on the internet. This will generate buzz and contributions, and with time, maybe a better extensions API than the one we're proposing here. But as it stands right now we need _some kind_ of extension API before we can do this. I promise we'll continue iterating on this (and keep pushing on this PR), but sadly our fork is going to be a hard requirement for fast iteration, at least on the short term.

Apologies and my deepest appreciation again.

MathieuDuponchelle:
@vmg, I'm a bit confused by this situation, I'll let @jgm tell us what he thinks I guess

MathieuDuponchelle:
Just a cheeky comment whlle I'm passing to check if there's anything new:

@kivikakk : 

> Took out the shared object searcher. I don't think library code should searching and loading objects dynamically at runtime. 

GStreamer begs to differ: https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstRegistry.html ;)

The correct argument here, IMHO is that the use case for user-side composing of plugins coming from multiple arbitrary sources is not as clear-cut for commonmark syntax rules as it is for creation of multimedia pipelines :)

jeroenooms:
My 2ct as a neutral bystander (and fan of cmark): is understandable that the commonmark team is  reluctant to drag in the complexity of the above extension framework. The simplicity of the current library makes it powerful, maintainable, and widely applicable.

Yet we really need some a way to make it easier to experiment with new features to further improve commonmark. For example, I personally maintain the cmark R bindings. Even though markdown is very popular in R, missing support for tables is a show stopper in our domain. For this reason R users still prefer the deprecated sundown stuff. There has been no progress on this front for too long.

An more modular, extensible version of `cmark` which allows for implementing and testing feature proposals could potentially be a great solution. However I really hope such a fork can be designed in a way that enables the commonmark team to selectively adopt feature proposals into the core commonmark specification and reference implementation. A split would be really unfortunate, the last thing we need is yet another markdown standard.

jgm:
@kivikakk @vmg I'm curious what the state of things is.  Is GitHub still working on its cmark fork?  You had talked about a time frame of weeks, so I expected we might hear something by now.

vmg:
@jgm: The comment you've just posted is rendered using CMark. Same as this one. We've implemented all the extensions we needed. We're now in the process of migrating (translating) all the historical comments so they can be rendered with CMark without being affected by any of the small differences between the new and the old parser.

We're not quite ready to OSS our fork but we'll be doing that soon, together with an announcement. Gotta finish migrating historical data first. 😄

jgm:
@vmg - That's exciting!

I'm curious how you're migrating historical data (does this involve translating to HTML and then using cmark to go back to commonmark?).

vmg:
We've implemented a new renderer for `sundown` that instead of outputting HTML, outputs CMark-compliant Markdown. It works very well in practice -- it's basically a "normalization pass" on the document.

jgm:
+++ Vicent Martí [Nov 25 16 06:37 ]:
>   We've implemented a new renderer for sundown that instead of outputting
>   HTML, outputs CMark-compliant Markdown. It works very well in practice
>   -- it's basically a "normalization pass" on the document.

Do you use cmark's commonmark renderer for this?

I ask because it's a surprisingly tough problem generating
commonmark that has the same meaning as some initial HTML.
It took quite a bit of work before cmark's renderer passed
round-trip tests on the test suite.  If you use something
else, I hope it is *very* well tested with round-trips.

There's always a tradeoff between guaranteeing correctness
and avoiding excessive backslash-escaping.  I've been
thinking about ways to get both, and it's a hard problem.

MathieuDuponchelle:
https://github.com/jgm/cmark/issues/131 for reference :)

jgm:
@vmg I just found a pathological case that cmark doesn't handle well (thanks to Martin Mitáš).
To generate it, use this program:
```c
#include <stdio.h>

int
main(int argc, char** argv)
{
    int i, j;

    for(i = 0; i < 10000; i++) {
        printf("x ");
        for(j = 0; j < i; j++)
            printf("`");
        printf("\n");
    }

    return 0;
}
```
I thought I should let you know since it's a potential security issue.

I'm looking into how to address this now.  The problem is obvious:  the backtick code span parser looks ahead for an appropriate closer and backtracks if one isn't found.  In this input, it goes back and forth over the whole input, scanning to the end, then coming back to the next potential starter.

A slightly different approach where we identify potential backtick delims early so we don't have to go over the input again would make sense.  (This would be a bit like what we do with emphasis and links now.)  Mitáš takes this approach in his md4c project.

vmg:
@kivikkak: Heads up for Monday!

jgm:
@vmg and @kivikakk - the easiest way to fix this would be to impose an upper limit on the length of backtick strings that can delimit code spans. A limit of (say) 100 should be enough for all practical purposes.  I don't like this kind of hard limit, and I'm thinking about some algorithmic improvements, but this would work for now....

jgm:
@vmg @kivikakk I have pushed some changes that help with the pathological case.
They also impose a 1000-character limit on backtick code span delimiters, which seems acceptable to me.

kivikakk:
@jgm great, thanks very much! Nice catch! :)

kivikakk:
We've now opened our fork at https://github.com/github/cmark — and I've just opened #166. :)

MathieuDuponchelle:
@kivikakk did you just attribute yourself authorship on my work ?? I am absolutely not OK with that ..

MathieuDuponchelle:
And by this I mean https://github.com/github/cmark/commit/2b0a0c4245782dbaddd414095df27192298d988c , the core and actual value of which was implemented by me.

kivikakk:
@MathieuDuponchelle: I didn't do that. If you're talking about https://github.com/github/cmark/commit/2b0a0c4245782dbaddd414095df27192298d988c, it's just a super-squashed commit; authorship is noted in the commit message.

MathieuDuponchelle:
@kivikakk that is not how git authorship works in the Open Source community, I really shouldn't have to explain that

MathieuDuponchelle:
Please "super-squash" what I actually did in a commit that authored by me, and split out your own additions in a separate one, that would be way more polite

kivikakk:
@MathieuDuponchelle: can you suggest a solution you'd be happy with? I'm not really interested in authorship itself as a concept, so if retagging the commit as yours would be sufficient, I'd most like to do that.

MathieuDuponchelle:
@kivikakk , see above, thanks, you don't get to decide whether your "lack of interest for authorship as a concept" extends to other people's work as well ;)

MathieuDuponchelle:
Same for the strikethrough extension, and the table extension, you seem to have reworked the latter but it would have been way more practical to actually see what your changes were if you had committed them separately, in any case I do have authorship on the original work, you didn't even mention it in the commit messages there ...

kivikakk:
@MathieuDuponchelle: I've split the commits up accordingly. It's not possible to separate the table and strikethrough extensions into separate commits because they're in entirely different files to begin with; the diff is simply minus `core-extensions.c`, plus `table.c`.

I hope this helps.

MathieuDuponchelle:
@kivikakk thanks.

I'd really like to know what your intentions are with this fork in the long term?

jeroenooms:
Thanks everyone, this is cool. You guys mention that this is already in use to render github comments. Is the syntax highlighting in github flavored markdown done as a preprocessing step or so?

vmg:
@MathieuDuponchelle: We don't have a specific goal with the fork. It's the code we use to render Markdown in GitHub and we wanted to OSS it so users can refer to it.

Of course, we would love it if @jgm were interested on reviewing the extensions we've implemented and consider merging those upstream! But he's already busy enough as it is, so we're happy to just start rendering all the documents we currently have as CommonMark + our unofficial extensions.

@jeroenooms: It's actually a post-processing step on the generated HTML, just like all the other small GitHub-specific features, like autolinking commits and user names. Check out http://github.com/jch/html-pipeline for an example of how to do this yourself.

jeroenooms:
OK. One final question: if no extensions are enabled in the parser, the output of this fork will be exactly identical to the original jgm/cmark code, correct? If we default to not enabling extensions, I should be able to use this fork as a drop-in replacement without any side effects?

vmg:
@jeroenooms: Yes! That's been the intention since day one and we've been trying _really hard_ to stick to that. Just this morning we've upstreamed the only slight modification we had performed on the CommonMark spec (a small change on the way links are parsed) and @jgm has merged it. So we should be 100% compatible.

By the way, I was going to followup by email, but yeah... There's the fork. Hopefully it'll be useful for you. :)

jgm:
+++ Vicent Martí [Dec 01 16 02:53 ]:

>   [3]@jeroenooms: It's actually a post-processing step on the generated
>   HTML, just like all the other small GitHub-specific features, like
>   autolinking commits and user names. Check out
>   [4]http://github.com/jch/html-pipeline for an example of how to do this
>   yourself.

It's pretty easy to use cmark's iterator interface to do
this kind of thing prior to rendering, instead of
postprocessing.

Here's an example of doing this in lcmark, my lua cmark
wrapper:

https://github.com/jgm/lcmark/blob/master/filters/highlight.lua

I would guess this approach would be more efficient than
postprocessing HTML, because you're crawling an efficient
representation of the document's nodes, and you don't need
to parse HTML.

jeroenooms:
OK this is all very cool stuff. I am going to release R bindings soon. Many R users have been waiting for a complete and portable markdown renderer which does not depend on haskell ;)

I do hope that the split will be temporary and we will eventually converge to combine efforts into a single "cmark" implementation.

jeroenooms:
Another question: in an embedded application, should `cmark_register_plugin(core_extensions_registration);` only be called once or each time a new parser is created?

jeroenooms:
@vmg I am getting the following compiler warnings:

```
c:/Rtools/mingw_64/bin/gcc  -I"C:/PROGRA~1/R/R-33~1.2/include" -DNDEBUG -Icmark -I.    -I"d:/Compiler/gcc-4.9.3/local330/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c cmark/iterator.c -o cmark/iterator.o
c:/Rtools/mingw_64/bin/gcc  -I"C:/PROGRA~1/R/R-33~1.2/include" -DNDEBUG -Icmark -I.    -I"d:/Compiler/gcc-4.9.3/local330/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c cmark/blocks.c -o cmark/blocks.o
cmark/blocks.c: In function 'cmark_manage_extensions_special_characters':
cmark/blocks.c:395:41: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
       unsigned char c = (unsigned char) (unsigned long) tmp_char->data;
                                         ^
c:/Rtools/mingw_64/bin/gcc  -I"C:/PROGRA~1/R/R-33~1.2/include" -DNDEBUG -Icmark -I.    -I"d:/Compiler/gcc-4.9.3/local330/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c cmark/inlines.c -o cmark/inlines.o
cmark/inlines.c: In function 'add_extensions_openers_bottom':
cmark/inlines.c:511:41: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
       unsigned char c = (unsigned char) (unsigned long) tmp_char->data;
                                         ^
cmark/inlines.c: In function 'get_extension_for_special_char':
cmark/inlines.c:525:45: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
       unsigned char tmp_c = (unsigned char) (unsigned long) tmp_char->data;
                                             ^
```

MathieuDuponchelle:
@jeroenooms , I don't know why they even kept this API, it's absolutely useless given there's no plugin loading mechanism anymore. What you want to do is instantiate extensions, and attach them on a per-parser basis. That way you can have multiple parsers with different extensions enabled in the same program. Here is an example of how that worked with a previous version of the extensions API:

https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_module.c#L462 and https://github.com/hotdoc/hotdoc/blob/master/hotdoc/parsers/cmark_include_extension.c

jeroenooms:
@MathieuDuponchelle hmmm OK. Well I need to get rid of these warnings to publish my bindings:


```
Found the following significant warnings:
  cmark/blocks.c:395:41: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  cmark/inlines.c:511:41: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  cmark/inlines.c:525:45: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
```

So it looks like on Win64, the pointer is not the same size as `unsigned long`. What would be the best way to fix this warning? Thanks for your amazing work on this btw!

MathieuDuponchelle:
Hm, the problem is similar to this: https://developer.gnome.org/glib/stable/glib-Type-Conversion-Macros.html and should be handled in the same way. I would suggest going with a solution along these lines as well.

MathieuDuponchelle:
Unfortunately, the code for this is a bit obfuscated in the configure.ac file of the glib, not readily copypastable I'm afraid @jeroenooms . The API could also be reworked to take a null-terminated char array fwiw.

nwellnhof:
@jeroenooms I'd either stop using linked lists which are unnecessarily complex, or make the payload a union of pointer and integer types. But all of this this doesn't belong here. @vmg Can you allow to open issues on github/cmark?

MathieuDuponchelle:
> @jeroenooms I'd either stop using linked lists which are unnecessarily complex

That's my second suggestion, it shouldn't be very hard to implement

> or make the payload a union of pointer and integer types

There shouldn't be any reason to do that, I don't think cmark is intended to be compilable on platforms with < 32 bits pointer sizes?

jeroenooms:
@nwellnhof @kivikakk @vmg it would be great to open issues / pr's for the fork. I'd like to add support for aligning tables ([example](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables)).

kivikakk:
I opened issues on the github/cmark fork; I forgot they might not be enabled by default.

We kept the plugin registration mechanism so e.g. the `cmark` binary can list extensions that are available, so the user can then enable them at their leisure:

``` shell
$ build/src/cmark --list-extensions
Available extensions:
table
strikethrough
autolink
tagfilter
$
```

(Per @MathieuDuponchelle's original design!) Without a registration mechanism, there's no way for the library to list them ahead of time. This is important for wrapping the library.

@jeroenooms: here's how we've done it in the Ruby `commonmarker` extension: https://github.com/gjtorikian/commonmarker/blob/d5c50a9238f5df210783ae6cbc41b66c02848be5/ext/commonmarker/commonmarker.c

We register the known core extensions once, on initialisation of the gem, and then provide the method `CommonMarker.extensions` to get a list of known extension names. The Ruby gem user can then supply an array of extension names when rendering, which are [attached to the parser](https://github.com/gjtorikian/commonmarker/blob/d5c50a9238f5df210783ae6cbc41b66c02848be5/ext/commonmarker/commonmarker.c#L136) or [supplied to `cmark_render_html`](https://github.com/gjtorikian/commonmarker/blob/d5c50a9238f5df210783ae6cbc41b66c02848be5/ext/commonmarker/commonmarker.c#L529-L537).

As for table alignment, it's already added! See the relevant part of the [extensions tests](https://github.com/github/cmark/blob/0e197aa8e00ad87405b058c3736e962ba44a67cb/test/extensions.txt#L172), and this example:

``` markdown
| longer table | headers to | demonstrate alignment |
| :-- | :-: | --: |
| a | b | c |
```

| longer table | headers to | demonstrate alignment |
| :-- | :-: | --: |
| a | b | c |

kivikakk:
@jeroenooms: could you try building this branch for your R bindings? https://github.com/github/cmark/tree/kivikakk/win32

Unfortunately I'm unable to get an exact build environment up that simulates yours (or perhaps that's CRAN's), but this might clear all the warnings.

kivikakk:
If there are any more issues with our fork, please feel free to open them at https://github.com/github/cmark/issues/new!

MathieuDuponchelle:
> (Per @MathieuDuponchelle's original design!) Without a registration mechanism, there's no way for the library to list them ahead of time. This is important for wrapping the library.

Given the reduction of scope in your branch, and the removal of *plugins*, the whole cmark_*plugin* API is simply useless. You could simply (and should, in order to provide an API that makes a minimal amount of sense), at cmark_init time, directly instantiate the extensions, *as you compile them along the rest of your sources*, instead of going through convoluted maneuvers in the cmark executable, involving (sic), a `#include "../extensions/core-extensions.h"`. Anyway ...

jeroenooms:
For those following this thread, I have posted an announcement of the R bindings here: https://ropensci.org/blog/blog/2016/12/02/commonmark

kivikakk:
@MathieuDuponchelle: (a) there is no more `cmark_init` in our fork, (b) it is designed so that the core cmark code can be compiled once and not further changed: it can be distributed as a library, and then you can compile or discover extensions however you like and then register them. You're right that I can ditch `cmark_plugin_*` functions; we should just directly allow addition of syntax extensions to the known registry.

`#include "../extensions/core-extensions.h"` is in the `main.c`, which is not part of the library. The program sources are only "main.c" itself, which is then linked with `libcmark` and `libcmarkextensions`. This works quite well.

I hope this has helped elucidate some of the design factors for you. Seeing as we've published our fork, I'll close this issue — please feel free to open issues on github/cmark if you have questions about our fork specifically!

@jeroenooms: the R package release notes look great! :)

MathieuDuponchelle:
Last remark, the design that was implemented by this pull request does not make it possible for a third party to implement a new inline extension and have it containable in, for example a table cell as implemented by the table extension. So much for compiling the core once ;)

PR #156, opened 2016-09-26
Use cmark_mem to free where used to alloc

kivikakk:
In a few places we don't use the `cmark_mem`'s `free` function when we used its `calloc`/`realloc`. This could bite someone using a custom allocator.


jgm:
Thanks!

PR #157, opened 2016-09-26
Fix memory leak in list parsing

kivikakk:
If `parse_list_marker` returns 1, but the second part of the `&&` clause
is false, we leak `data` here.


jgm:
Great, thanks.  I introduced that by accident with the CommonMark 0.26 revisions.

PR #158, opened 2016-09-27
One more cmark_mem use

kivikakk:
I missed one in #156; `print_document` itself frees a string allocated by the document's allocator via the renderer.


jgm:
Great, thanks

PR #160, opened 2016-10-10
Fix NUL-LF sequence lex

### LINKED ISSUE #159: Wrong parsing of '\u0000' inside a fenced code block
The bash command 

```
$ printf '```\n\u0000\n```\n' | cmark
```
results in

```
<pre><code>�```
</code></pre>
```

but I would expect this (i.e. the bacticks should not be present in the output):

```
<pre><code>�
</code></pre>
```

kivikakk:
Fixes #159.

If we hit a NUL, we advance past it. But then we advance past any LF or CR-LF, assuming we're at the end of a line and just need to push past the newline that broke us from the loop.

### DIFF
--- a/api_test/main.c
+++ b/api_test/main.c
@@ -732,6 +732,13 @@ static void utf8(test_batch_runner *runner) {
       string_with_null, sizeof(string_with_null) - 1, CMARK_OPT_DEFAULT);
   STR_EQ(runner, html, "<p>((((" UTF8_REPL "))))</p>\n", "utf8 with U+0000");
   free(html);
+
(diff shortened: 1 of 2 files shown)


jgm:
Thanks!  I may make some minor changes to this.

kivikakk:
Looks good! Although the spacing seems to have gone slightly awry.

jgm:
I think I fixed the spacing in the next commit.

(I updated 'make format' so it doesn't reimport imports,
which was causing problems.)

+++ Yuki Izumi [Oct 11 16 07:20 ]:

PR #161, opened 2016-11-04
Fix for non-matching entities

kivikakk:
Users were noticing that entities like `&larr;` and `&rarr;` were not being converted correctly (and were getting rendered as `&amp;larr;`, etc.).  A slight bug in the string comparison in `S_lookup` was the cause.  See commit message of a63a3c7 for more details.


jgm:
Great, thanks.

PR #162, opened 2016-11-09
Fix CMAKE_C_VISIBILITY_PRESET for cmake versions greater than 1.8 (e.g. 3.6.2)

hughbe:
This lets us build swift-cmark on Windows, using `clang-cl`

This was an oversight: CMAKE_C_VISIBILITY_PRESET is only available on CMAKE 1.8 or greater. However, the current check doesn't work for a version such as CMAKE 3.6.2, for example.

If this is not added, we get an error: `error: unknown argument: '-fvisibility=hidden'`

```bash
mkdir "C:/Users/hughb/Documents/GitHub/my-swift/build/Ninja-DebugAssert/cmark-windows-amd64"
pushd "C:/Users/hughb/Documents/GitHub/my-swift/build/Ninja-DebugAssert/cmark-windows-amd64"
cmake -G Ninja -DCMAKE_C_COMPILER="C:/Program Files/LLVM/msbuild-bin/cl.exe" -DCMAKE_CXX_COMPILER="C:/Program Files/LLVM/msbuild-bin/cl.exe" -DCMAKE_CXX_FLAGS="-Xclang -std=c++14" "C:/Users/hughb/Documents/GitHub/my-swift/cmark"
popd
cmake --build "C:/Users/hughb/Documents/GitHub/my-swift/build/Ninja-DebugAssert/cmark-windows-amd64/" -- -j6 all
```

/cc @jgm @kivikakk


jgm:
Thanks!

kainjow:
Side note: wouldn't it be better to use `VERSION_GREATER` here? It's available for [2.8.9](https://cmake.org/cmake/help/v2.8.9/cmake.html).

jgm:
+++ Kevin Wojniak [Nov 09 16 09:24 ]:
>   Side note: wouldn't it be better to use VERSION_GREATER here? It's
>   available for [1]2.8.9.

I didn't even know about VERSION_GREATER, but that does
sound right.  Feel free to submit a PR!

PR #163, opened 2016-11-09
Remove unnecessary cmake_minimum_required

kainjow:
Top-level CMakeLists already includes this, and uses a different version number anyways


jgm:
Thanks! I did this in a separate commit because of conflicts.

PR #168, opened 2016-12-03
Update CaseFolding to latest

kainjow:
1. Downloaded CaseFolding.txt from http://unicode.org/Public/UCD/latest/ucd/CaseFolding.txt
2. Deleted src/case_fold_switch.inc
3. Ran `make src/case_fold_switch.inc`


jgm:
Thanks!

PR #169, opened 2016-12-09
More sourcepos!

MathieuDuponchelle:
Found time to work a bit more on https://github.com/jgm/cmark/issues/26 , these commits solely implement the definition of a gap-free source map with no empty extents, parsed in the same pass as the nodes.

A few things to discuss:

* Ownership of line-termination characters and trailing whitespace at the end of blocks, I make them belong to the root
* Ownership of reference definitions, which I also attribute to the root, as there's no proper AST node for them
* Performance, my initial implementation for this approach doubled make bench time (erg), I now observe a 25 % difference, which seems reasonable to me, not sure how much can be ironed out still, the task is inherently a bit complex, due to potential discontinuities between the n extents constituting a block. Either we consider this a reasonable enough performance hit and hope we find more smart ways to reduce it, or we make it conditional to the SOURCEPOS option as currently.

Things that are in my opinion out of scope and can be discussed later:

* Actual API for this
* "Reverse source map", ie node to extents

To test this, just enable the call to "print_parser_source_map" in blocks.c:finalize_document , example output for the case that made me revise my approach:

```
>     code
>     more code
```

```
0:1 - block_quote (0x9485f0)
1:2 - block_quote (0x9485f0)
2:6 - code_block (0x948740)
6:14 - code_block (0x948740)
14:15 - block_quote (0x9485f0)
15:16 - block_quote (0x9485f0)
16:20 - code_block (0x948740)
20:30 - code_block (0x948740)
```

I'm sure there's more to say, but let's keep this short :)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,6 +18,7 @@ set(HEADERS
   houdini.h
   cmark_ctype.h
   render.h
+  source_map.h
   )
 set(LIBRARY_SOURCES
   cmark.c
@@ -40,6 +41,7 @@ set(LIBRARY_SOURCES
   houdini_html_e.c
   houdini_html_u.c
   cmark_ctype.c
+  source_map.c
   ${HEADERS}
   )
 
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -867,8 +867,8 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
     indented = parser->indent >= CODE_INDENT;
 
     if (!indented && peek_at(input, parser->first_nonspace) == '>') {
-
-      bufsize_t blockquote_startpos = parser->first_nonspace;
+      *container = add_child(parser, *container, CMARK_NODE_BLOCK_QUOTE,
+                             parser->first_nonspace + 1);
 
       S_advance_offset(parser, input,
                        parser->first_nonspace + 1 - parser->offset, false);
@@ -876,20 +876,17 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
       if (S_is_space_or_tab(peek_at(input, parser->offset))) {
         S_advance_offset(parser, input, 1, true);
       }
-      *container = add_child(parser, *container, CMARK_NODE_BLOCK_QUOTE,
-                             blockquote_startpos + 1);
 
     } else if (!indented && (matched = scan_atx_heading_start(
                                  input, parser->first_nonspace))) {
       bufsize_t hashpos;
       int level = 0;
-      bufsize_t heading_startpos = parser->first_nonspace;
 
+      *container = add_child(parser, *container, CMARK_NODE_HEADING,
+                             parser->first_nonspace + 1);
       S_advance_offset(parser, input,
                        parser->first_nonspace + matched - parser->offset,
                        false);
-      *container = add_child(parser, *container, CMARK_NODE_HEADING,
-                             heading_startpos + 1);
 
       hashpos = cmark_chunk_strchr(input, '#', parser->first_nonspace);
 
@@ -936,9 +933,9 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
                !(cont_type == CMARK_NODE_PARAGRAPH && !all_matched) &&
                (matched = scan_thematic_break(input, parser->first_nonspace))) {
       // it's only now that we know the line is not part of a setext heading:
+      S_advance_offset(parser, input, input->len - 1 - parser->offset, false);
       *container = add_child(parser, *container, CMARK_NODE_THEMATIC_BREAK,
                              parser->first_nonspace + 1);
-      S_advance_offset(parser, input, input->len - 1 - parser->offset, false);
     } else if ((!indented || cont_type == CMARK_NODE_LIST) &&
                (matched = parse_list_marker(
                     parser->mem, input, parser->first_nonspace,
@@ -946,7 +943,22 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
 
       // Note that we can have new list items starting with >= 4
       // spaces indent, as long as the list container is still open.
+      cmark_node *list = NULL;
+      cmark_node *item = NULL;
       int i = 0;
+
+      if (cont_type != CMARK_NODE_LIST ||
+          !lists_match(&((*container)->as.list), data)) {
+        *container = add_child(parser, *container, CMARK_NODE_LIST,
+                               parser->first_nonspace + 1);
+        list = *container;
+
+      }
+
+      // add the list item
+      *container = add_child(parser, *container, CMARK_NODE_ITEM,
+                             parser->first_nonspace + 1);
+      item = *container;
 
       // compute padding:
       S_advance_offset(parser, input,
@@ -982,22 +994,14 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
 
       data->marker_offset = parser->indent;
 
-      if (cont_type != CMARK_NODE_LIST ||
-          !lists_match(&((*container)->as.list), data)) {
-        *container = add_child(parser, *container, CMARK_NODE_LIST,
-                               parser->first_nonspace + 1);
-
-        memcpy(&((*container)->as.list), data, sizeof(*data));
-      }
-
-      // add the list item
-      *container = add_child(parser, *container, CMARK_NODE_ITEM,
-                             parser->first_nonspace + 1);
       /* TODO: static */
-      memcpy(&((*container)->as.list), data, sizeof(*data));
+      if (list)
+        memcpy(&(list->as.list), data, sizeof(*data));
+      if (item)
+        memcpy(&(item->as.list), data, sizeof(*data));
+
       parser->mem->free(data);
     } else if (indented && !maybe_lazy && !parser->blank) {
-      S_advance_offset(parser, input, CODE_INDENT, true);
       *container = add_child(parser, *container, CMARK_NODE_CODE_BLOCK,
                              parser->offset + 1);
       (*container)->as.code.fenced = false;
@@ -1006,6 +1010,7 @@ static void open_new_blocks(cmark_parser *parser, cmark_node **container,
       (*container)->as.code.fence_offset = 0;
       (*container)->as.code.info = cmark_chunk_literal("");
 
+      S_advance_offset(parser, input, CODE_INDENT, true);
     } else {
       break;
     }
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -28,6 +28,10 @@
 #define MIN(x, y) ((x < y) ? x : y)
 #endif
 
+#ifndef MAX
+#define MAX(x, y) ((x > y) ? x : y)
+#endif
+
 #define peek_at(i, n) (i)->data[n]
 
 static bool S_last_line_blank(const cmark_node *node) {
@@ -93,6 +97,7 @@ cmark_parser *cmark_parser_new_with_mem(int options, cmark_mem *mem) {
   parser->root = document;
   parser->current = document;
   parser->line_number = 0;
+  parser->line_offset = 0;
   parser->offset = 0;
   parser->column = 0;
   parser->first_nonspace = 0;
@@ -103,6 +108,7 @@ cmark_parser *cmark_parser_new_with_mem(int options, cmark_mem *mem) {
   parser->last_line_length = 0;
   parser->options = options;
   parser->last_buffer_ended_with_cr = false;
+  parser->source_map = source_map_new(mem);
 
   return parser;
 }
@@ -116,6 +122,7 @@ void cmark_parser_free(cmark_parser *parser) {
   cmark_mem *mem = parser->mem;
   cmark_strbuf_free(&parser->curline);
   cmark_strbuf_free(&parser->linebuf);
+  source_map_free(parser->source_map);
   cmark_reference_map_free(parser->refmap);
   mem->free(parser);
 }
@@ -255,10 +262,13 @@ static cmark_node *finalize(cmark_parser *parser, cmark_node *b) {
 
   switch (S_type(b)) {
   case CMARK_NODE_PARAGRAPH:
+    source_map_start_cursor(parser->source_map, parser->last_paragraph_extent);
     while (cmark_strbuf_at(node_content, 0) == '[' &&
            (pos = cmark_parse_reference_inline(parser->mem, node_content,
-                                               parser->refmap))) {
-
+                                               parser->refmap, parser->root,
+                                               parser->source_map))) {
+      parser->last_paragraph_extent = parser->source_map->cursor;
+      source_map_start_cursor(parser->source_map, parser->last_paragraph_extent);
       cmark_strbuf_drop(node_content, pos);
     }
     if (is_blank(node_content, 0)) {
@@ -266,7 +276,6 @@ static cmark_node *finalize(cmark_parser *parser, cmark_node *b) {
       cmark_node_free(b);
     }
     break;
-
   case CMARK_NODE_CODE_BLOCK:
     if (!b->as.code.fenced) { // indented code
       remove_trailing_blank_lines(node_content);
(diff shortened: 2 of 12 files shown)


MathieuDuponchelle:
heh, some leaks, no invalid reads, I'll look at that tomorrow that shouldn't be a blocker for reviewing :)

jgm:
This looks really promising.  A couple thoughts:

- Would it be hard to adjust the code so that adjacent blocks get consolidated?
  E.g. instead of 
  ```
  4446:4450 - code_block (0x7fa913e068c0)
  4450:4456 - code_block (0x7fa913e068c0)
  4456:4457 - code_block (0x7fa913e068c0)
  ```
  we'd get
  ```
   4446:4457 - code block (0x7fa913e068c0)
   ```
  This shouldn't be too hard; just a matter of examining the previous node before adding a new one, right?

- As a way of demonstrating this, it would be nice to add a "syntax highlighter" that simply reads the source text and spits out literal (HTML-escaped) text, with HTML span or div tags inserted when needed according to the source map.  "Look, 100% accurate, fast CommonMark syntax highlighting!"

jgm:
A note on syntax highlighting:  the current system will make it easy to highlight the markers for a block quote, but the paragraph inside the block quote will be highlighted as a regular paragraph, not e.g. put in a special color.  I don't know if that's always going to be what is wanted, but it's something to think about.

I think a lot of people who care about syntax highlighting may want, e.g. to highlight the TEXT of a header, not just the opening and closing `#` marks.  And it's hard to see how to do that with this system, since there's no explicit marker for the end of a block.

Here's an idea. Instead of a linked list of nodes for extents, which brings in lots of memory allocation (that is probably what is responsible for the performance drop), why not have an array of structs, of length equal to the length of the source document?  (It could be allocated once at the beginning.)

Each element of the array would contain:
- a pointer to a node
- an indication BEGIN, END, or CONTINUED

Then we'd really have all the information that could conceivably be needed.  This array might end up taking more memory than your linked list, but it shouldn't be more than 10x the size of the source document.

jgm:
Further thought:  why not have overlapping extents?
E.g. in your example above
```
0:30 - block_quote (0x9485f0)
2:14 - code_block (0x948740)
16:20 - code_block (0x948740)
20:30 - code_block (0x948740)
```
since after all, ALL of this is part of the block quote; only the code block occupies discontinuous positions.

MathieuDuponchelle:
Will answer these one by one:

> Would it be hard to adjust the code so that adjacent blocks get consolidated?

it would not, but these adjacent blocks are deliberate: the goal of this step is to define all the extents that make up a block, the goal of the next one will be to offer a meaning to them, either by tagging the extent, or assigning it to a node field:

```
* a
```

In that case, I want to know that not only "*" makes up an item, but also that it represents a list item marker (this would adequately solve https://github.com/jgm/cmark/issues/125 by the way :D).

Most of these block extents are created when advance_offset is called, we do have that semantic information at that time. It might make sense to group advance_offset calls in certain places, but I think most extents are currently divided to my liking.

> As a way of demonstrating this, it would be nice to add a "syntax highlighter"  [...]

Sure, that is a fun and easy way of putting this to use, shouldn't be too hard I can figure out something :)

> I think a lot of people who care about syntax highlighting may want, e.g. to highlight the TEXT of a header, not just the opening and closing # marks. And it's hard to see how to do that with this system, since there's no explicit marker for the end of a block.

So that's what I want to address with the "reverse source map", which will make this trivial, and I think consolidating and overlapping extents go directly against this :)

> Here's an idea. Instead of a linked list of nodes for extents, which brings in lots of memory allocation (that is probably what is responsible for the performance drop), why not have an array of structs [...]

This would be the real performance killer, as we'd end up with long for loops all over the place: when advancing the offset by n, you'd have to iterate the extents list by n as well. Most of the performance hit we were taking was because I was more correctness than performance-minded during development, and pruned 0-length extents in a final pass, iterating over all the created extents (the algorithm allows for 0-length block extents to be created, in order for inline extents to have a start extent to be "spliced in" from. This was with considerably less extents that one per character, so I can only imagine the performance impact your solution would have :) Note that memory allocators are very smart these days (which is why I think the ubiquitous mem thing is completely useless, but I digress :P)

> Well, I suppose one problem is that often two nodes will begin or end on the same character position (e.g. a list and a list item might begin at the same position, and a block quote and a paragraph it contains might end at the same position).

Hm yes, though that wouldn't be the argument I would use here, as I don't think we want overlapping extents, see my next answer :)

> Further thought: why not have overlapping extents?

Because of the first answer here: I want extents that have semantic meaning, overlapping extents would go against that. In the next step of that work, we will make it trivial to actually list all the extents a given node is made of, including if desirable its children nodes. That can completely be done as a post-processing step on a per-client basis, and preserves this non-overlapping list of extents for usages such as "passthrough" rendering :)

I hope I've answered your questions adequately, last time around I completely ditched my approach because you were very right, but this time I'm quite sure the design is adequate :)

jgm:
+++ Mathieu Duponchelle [Dec 09 16 04:11 ]:
>   Because of the first answer here: I want extents that have semantic
>   meaning, overlapping extents would go against that. In the next step of
>   that work, we will make it trivial to actually list all the extents a
>   given node is made of, including if desirable its children nodes. That
>   can completely be done as a post-processing step on a per-client basis,
>   and preserves this non-overlapping list of extents for usages such as
>   "passthrough" rendering :)
>
>   I hope I've answered your questions adequately, last time around I
>   completely ditched my approach because you were very right, but this
>   time I'm quite sure the design is adequate :)

So, with a syntax highlighter, I'd imagine it would be more
sensible to iterate through the source map directly rather
than the node tree.

And, at least in HTML, what a syntax highlighter needs to do
is put in tokens at the beginnings and ends of things.

I still don't see how this is going to work since your
sourcemap representation doesn't indicate where the ends
of blocks are.  Can you explain that?

MathieuDuponchelle:
> So, with a syntax highlighter, I'd imagine it would be more sensible to iterate through the source map directly rather than the node tree.

Yes.

> And, at least in HTML, what a syntax highlighter needs to do is put in tokens at the beginnings and ends of things.
> I still don't see how this is going to work since your sourcemap representation doesn't indicate where the ends of blocks are. Can you explain that?

Of course, it's very simple:

Given this input:

```
> a
```

What we want to highlight here is ">" and " " as part of the block quote, and "a" as inline text. No proble whatsoever ?

jgm:
+++ Mathieu Duponchelle [Dec 09 16 07:05 ]:
>   Of course, it's very simple:
>
>   Given this input:
>> a
>
>   What we want to highlight here is ">" and " " as part of the block
>   quote, and "a" as inline text. No proble whatsoever ?

OK, yes, if you only want the > to appear in the special
color for blockquotes.  But normally in Markdown syntax
highlighting more is done.  The text of a header (not just
the opening #) might appear in boldface.  A blockquote
might be a special color or have some other special
formatting -- not just the opening `>` character but the
text too.

MathieuDuponchelle:
Still not a problem though, I'll make a simple demo later I guess

MathieuDuponchelle:
So @jgm, here's a simple example, I hope you don't mind the colors they're generated procedurally from a set of random colors, basically each extent is wrapped in a pre, this pre is given the classes of its node and its ancestors, and then a very dumb script mixes the colors to have text belonging to a heading render differently than text belonging to a paragraph, doing this client-side is terribly inefficient but I was too lazy to implement that in C :)

https://people.collabora.com/~meh/silly_highlighter/

jgm:
OK, I see.  Right, once you have a pointer to a node, you
have all the ancestors and can just include those classes.

MathieuDuponchelle:
@jgm, pretty much yep :)

I'm pretty excited about this work, opens up a whole bunch of applications for others to build on, I don't intend to work on a serious syntax highlighter, but I think that shows it's easily feasible, and the result will be accurate :)

MathieuDuponchelle:
Well, I've defined an API, and an approach for assigning types to source extents, plus I wanted to write my refactoring tool in python so I completed the python wrappers in a separate commit, along with a pretty extensive test coverage.

Memory management in the bindings is appropriate, except for the parser's awkward lifecycle with respect to its root node, I had a commit in my extensions' branch to address that.

@jgm, I'd really appreciate if we could go ahead with at least the first two commits here, I really don't want to end up with the same situation as last time around :)

jgm:
Looks good!  I'll be interesting to see what can be done with this.

jgm:
I'm getting the following warning (with clang):
```
/Users/jgm/src/cmark/src/source_map.c:199:40: warning: /Users/jgm/src/cmark/src/source_map.c:199:40: warning: format specifies type 'unsigned long' but the argument has type 'uint64_t'
      (aka 'unsigned long long') [-Wformat]
format specifies type 'unsigned long' but the argument has type 'uint64_t'
      (aka 'unsigned long long') [-Wformat]
    printf ("%lu:%lu - %s, %s (%p)\n", tmp->start, tmp->stop,
             ~~~                       ^~~~~~~~~~
             %llu
/Users/jgm/src/cmark/src/source_map.c:199:52: warning: format specifies type 'unsigned long' but the argument has type 'uint64_t'
      (aka 'unsigned long long') [-Wformat]
    printf ("%lu:%lu - %s, %s (%p)\n", tmp->start, tmp->stop,
                 ~~~                               ^~~~~~~~~
                 %llu
    printf ("%lu:%lu - %s, %s (%p)\n", tmp->start, tmp->stop,
             ~~~                       ^~~~~~~~~~
             %llu
/Users/jgm/src/cmark/src/source_map.c:199:52: warning: format specifies type 'unsigned long' but the argument has type 'uint64_t'
      (aka 'unsigned long long') [-Wformat]
    printf ("%lu:%lu - %s, %s (%p)\n", tmp->start, tmp->stop,
                 ~~~                               ^~~~~~~~~
```
How would you suggest fixing?

jgm:
With python bindings I get:
```
10/10 Test: python3_bindings
Command: "/usr/local/bin/python3" "/Users/jgm/src/cmark/test/test_cmark.py" "/Users/jgm/src/cmark/build/testdir/../src"
Directory: /Users/jgm/src/cmark/build/testdir
"python3_bindings" start time: Dec 20 16:44 EST
Output:
----------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jgm/src/cmark/test/test_cmark.py", line 12, in <module>
    from wrapper import *
  File "/Users/jgm/src/cmark/test/../wrappers/wrapper.py", line 14, in <module>
    libc = CDLL('libc.so.6')
  File "/usr/local/Cellar/python3/3.5.2/Frameworks/Python.framework/Versions/3.5/lib/python3.5/ctypes/__init__.py", line 347, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(libc.so.6, 6): image not found
<end of output>
```
Not sure how travis passed this with these warnings and test failures.  (Ah, I see, the python binding tests are being skipped. Why?)

foonathan:
Regarding the warnings: the correct way of solving it would be to use the appropriate macro: http://en.cppreference.com/w/cpp/types/integer#Format_constants_for_the_std::fprintf_family_of_functions

Here `PRlu64`, so `"%" PRlu64 ":%" PRlu64 " - %s, %s (%p)\n"` for example.

MathieuDuponchelle:
You rock, thanks!

I know how to fix these issues, I'll prepare some patches tomorrow. @foonathan , I didn't know about these macros, but a priori we don't need them, just need to use %llu? Not sure why I didn't see these warnings, sorry about that. Regarding the loading of the libc .so, we can just remove it altogether, I was loading it before exposing the cmark_default_mem_free API, which is a bit awkward but does the job, as we don't expose or use the memory API in the python bindings.

foonathan:
@MathieuDuponchelle Ah, misinterpreted the warnings.
I thought clang would warn about the fact that you weren't using the macros, but it just says that you should use `llu`.

The macros are there because `std::uint64_t` might not be `unsigned long long` on every platform, so using `llu` is technically not portable.

MathieuDuponchelle:
@foonathan yep indeed, if you feel like doing the patch for that please don't hesitate, a bit busy right now but having master build with warnings isn't great :)

foonathan:
I can open a PR tomorrow. Should I change it to `llu` or all instances to the (more correct but ugly) macro?

MathieuDuponchelle:
The reason why @jgm was seeing this and not me is: http://stackoverflow.com/questions/14073877/printf-and-llu-vs-lu-on-os-x , iirc he is working on OSX, and I'm on a 64-bit linux. We don't really have a choice, and should use these macros IMHO

nwellnhof:
Do we really need a 64-bit type here? The type for internal string sizes and offsets is `bufsize_t` which is currently typedef'd to `int`, so strings are limited to 2 GB anyway.

MathieuDuponchelle:
> Do we really need a 64-bit type here? The type for internal string sizes and offsets is bufsize_t which is currently typedef'd to int, so strings are limited to 2 GB anyway.

We do, when using the streaming API you could feed larger than 2 GB files in there.

nwellnhof:
> We do, when using the streaming API you could feed larger than 2 GB files in there.

That's right, but blocks larger than 2 GB also cause a `bufsize_t` overflow making `cmark_strbuf_grow` abort. So I don't see why we shouldn't impose a limit on the maximum number of total bytes passed via the streaming API. (As already discussed, this should ideally be handled by returning error codes.)

Even without such a limit, `size_t` would probably be a better choice. On typical 32-bit platforms, it's impossible to handle files larger than 4 GB anyway, because memory will run out.

MathieuDuponchelle:
@nwellnhof , I think you didn't understand what this offset was, it's an offset in bytes from the start of the document, not the start of the block!

nwellnhof:
I'm aware of that. I just don't think it makes much sense to support documents larger than 2 GB without supporting blocks larger than 2 GB. The idea behind `bufsize_t` is that even if we want to support multi-gigabyte documents at some point, we only have to change the typedef.

jgm:
I'm sorry, I've had to revert this commit (as well as the revisions by @nwellnhof that depend on it).

The performance hit was far too big:  in my measurements, cmark ran almost twice as slow after your change (2.39s vs 1.37s).

If this feature is to be merged into cmark, it needs to be made conditional on an option flag so it doesn't destroy performance for people who don't need the feature.

MathieuDuponchelle:
@jgm hm OK, I think you've also accidentally the whole python bindings ;)

As for performance, I was observing a 25% performance hit, as I mentioned in the first comment, what you describe here is more like 75 % , which is indeed way too high, however it might not have been strictly needed to hastily revert this, I am absolutely not opposed to making this conditional, that's what I was asking about initially :)

Anyway, I have got a bunch of commits on top of that that result in a formatting script that can reindent code with minimal escaping, no style changes and preserve reference links, which I've pretty thoroughly tested on the 610 tests of the spec, by reindenting with width [1:80] and checking that the result still yielded the same AST, minus softbreaks, you were wondering what this work could lead to and I think this will make it pretty clear :)

I'll have a new pull request ready pretty soon.

jgm:
Great, I'll be happy to put all of this back with
conditionality.

+++ Mathieu Duponchelle [Jan 04 17 05:06 ]:

MathieuDuponchelle:
OK so just for the record, here are the results from make bench on my end:

with master:

```
{ for x in `seq 1 10` ; do \
	/usr/bin/env time -p build/src/cmark </dev/null >/dev/null ; \
	/usr/bin/env time -p build/src/cmark bench/benchinput.md >/dev/null ; \
	done \
} 2>&1  | grep 'real' | awk '{print $2}' | python3 'bench/stats.py'
mean = 0.6620, median = 0.6500, stdev = 0.0235
```

and with the source map code:

```
{ for x in `seq 1 10` ; do \
	/usr/bin/env time -p build/src/cmark </dev/null >/dev/null ; \
	/usr/bin/env time -p build/src/cmark bench/benchinput.md >/dev/null ; \
	done \
} 2>&1  | grep 'real' | awk '{print $2}' | python3 'bench/stats.py'
mean = 0.9940, median = 0.9900, stdev = 0.0143
```

Anyway, even 10 % would arguably be too much for a feature most won't need

jgm:
+++ Mathieu Duponchelle [Jan 04 17 12:01 ]:
>   Anyway, even 10 % would arguably be too much for a feature most won't
>   need

Yes, I think that's right.  (Somehow I had mentally read
your original message as 2.5% instead of 25%, or I would
have insisted on putting it behind a flag from the
beginning.)

MathieuDuponchelle:
@jgm right, makes sense. Note that I am very much around these days, so had you pinged me about this I'd have addressed this pretty fast without needing to revert anything :)

ericpruitt:
Hey @jgm and @MathieuDuponchelle, are there any plans to merge an updated version of this patch?

PR #170, opened 2016-12-09
Correctly initialize chunk in S_process_line

nwellnhof:
The `alloc` member wasn't initialized.

This also allows to add an assertion in `chunk_rtrim` which doesn't
work for alloced chunks.


jgm:
Thanks!

PR #172, opened 2016-12-22
Fix format specifier for uint64_t

foonathan:
As discussed in #169.

PR #174, opened 2016-12-23
Change types for source map offsets

nwellnhof:
As proposed in #169, this changes the types for source map offsets to `size_t` externally and `bufsize_t` internally. An additional check is added in `S_parser_feed` to make sure that the total input document size never exceeds an internal limit (currently about 2 GB). Obsoletes #172.

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -96,6 +96,7 @@ cmark_parser *cmark_parser_new_with_mem(int options, cmark_mem *mem) {
   parser->refmap = cmark_reference_map_new(mem);
   parser->root = document;
   parser->current = document;
+  parser->total_bytes = 0;
   parser->line_number = 0;
   parser->line_offset = 0;
   parser->offset = 0;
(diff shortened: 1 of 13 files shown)


src/blocks.c
```diff
@@ -550,6 +551,16 @@ static void S_parser_feed(cmark_parser *parser, const unsigned char *buffer,
   const unsigned char *skipped;
   static const uint8_t repl[] = {239, 191, 189};
 
+  // Limit maximum document size to BUFSIZE_MAX. This makes sure that we
+  // never create strbufs larger than BUFSIZE_MAX. Unfortunately, the
+  // public API doesn't have an error reporting mechanism, so all we can
+  // do is to abort.
+  if (len > (size_t)(BUFSIZE_MAX - parser->total_bytes)) {
+    fprintf(stderr, "Maximum cmark document size exceeded\n");
+    abort();
```
MathieuDuponchelle: Libraries should never call abort

MathieuDuponchelle:
That is not the correct solution IMHO, you can absolutely return when the string buffer for a block exceeds the maximum allowed size, and still continue to allow larger than 2 gigabytes input, this looks to me like an arbitrary restriction, not a necessary one.

MathieuDuponchelle:
Also, I really don't think you should ever call abort() in a ibrary that may be used by other programs or (even worse), in higher level libraries.

nwellnhof:
> Also, I really don't think you should ever call abort() in a ibrary that may be used by other programs or (even worse), in higher level libraries.

Yes, I agree. But IMO, the best solution is to change the API to return error codes instead of making internal types 64 bit wide and hoping that an overflow will never happen. Commit 36dca95 goes in that direction because it makes it easier to check for overflow of some internal integers. The reasoning is that if the input document is smaller than N bytes, all string buffers that are created during parsing are guaranteed to be smaller than N bytes, too. So we only have to check in one place and not on every strbuf operation like append. (Although ultimately, we should check such function calls anyway because I'd really like to see proper error returns for functions that may fail to allocate memory.)

Now it's easy to make `cmark_parser_finish` return NULL on error and add a function like `cmark_parser_get_error()` that returns an error code. The only downside is that documents larger than 2 GB can't be handled but this shouldn't be a serious concern.

jgm:
If it's easy to avoid the abort by returning null and providing an additional function to return an error code, as @nwellnhof suggests, then I think that's what we should do.

PR #179, opened 2017-01-04
CMake impovement

winterheart:
- Added support for built-in ${LIB_SIFFIX} feature, so CMake will automatically generate correct path for 32/64-bit environment.
- Making shared and static libraries optional to compile and install via -DCMARK_SHARED=ON/OFF and -DCMARK_STATIC=ON/OFF options.  By default both targets are set as enabled (previous behavior).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,9 +22,11 @@ set(PROJECT_VERSION_PATCH 1)
 set(PROJECT_VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH} )
 
 option(CMARK_TESTS "Build cmark tests and enable testing" ON)
+option(CMARK_STATIC "Build static libcmark library" ON)
+option(CMARK_SHARED "Build shared libcmark library" ON)
 
 add_subdirectory(src)
-if(CMARK_TESTS)
+if(CMARK_TESTS AND CMARK_SHARED)
   add_subdirectory(api_test)
 endif()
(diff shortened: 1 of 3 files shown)


kainjow:
Instead of custom CMAKE_SHARED and CMAKE_STATIC, why not use the built-in BUILD_SHARED_LIBS variable, supported in [2.8.12](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:add_library)?

> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is true.

winterheart:
@jgm 

1. Unfortunately, I have not found any references in CMake documentation, when LIB_SUFFIX was added. I suppose this feature already there since creation of CMake. At least 2.8.9 is OK.
2. Done.

@kainjow The main reason to use two options is make both of them optional. For example, I need in my system only cmark executable. Your proposal will make one of them as required.

jgm:
Thanks!

kainjow:
@winterheart BUILD_SHARED_LIBS is optional. If you don't set it, CMake builds static libraries. If you set it to true, then shared libraries are built. For example, with this CMakeLists.txt file:

    cmake_minimum_required(VERSION 2.8.9)
    project(cmaktest)
    add_library(cmaketest code.c)

If I compile it with `cmake .` then I get static libraries. If I compile with `cmake -DBUILD_SHARED_LIBS=true .` then I get shared libraries. You can use the variable in `if` statements like normal:

    if(BUILD_SHARED_LIBS)
        ...
    endif()

PR #182, opened 2017-01-07
Fixes for the LaTeX renderer

Doeme:
The first fix corrects ligature handling with dashes, (\- are manual hyphenations), the second fixes a double-output of urls.

### DIFF
--- a/src/latex.c
+++ b/src/latex.c
@@ -391,7 +391,7 @@ static int S_render_node(cmark_renderer *renderer, cmark_node *node,
         LIT("\\url{");
         OUT(url, false, URL);
         LIT("}");
-        return 0;
+        return 0; // Don't process further nodes to avoid double-rendering artefacts
       case EMAIL_AUTOLINK:
         LIT("\\href{");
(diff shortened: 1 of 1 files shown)


src/latex.c
```diff
@@ -389,7 +389,6 @@ static int S_render_node(cmark_renderer *renderer, cmark_node *node,
       switch (get_link_type(node)) {
       case URL_AUTOLINK:
         LIT("\\url{");
-        OUT(url, false, URL);
```
jgm: This fixes the double rendering, but we lose proper escaping for the URL.  (That is what the URL in OUT accomplishes.)  The correct fix would be to keep this line and somehow suppress the output of children of this node.  To do this, return 0 (that stops processing of the contents of the node):
```
-        break;
+       LIT("}");
+       return 0; // don't process contents or we get double render
```

Doeme: Ah, ok.
I've improved this in 1c0d4749451c. Thanks for looking over it!

PR #183, opened 2017-01-19
Fix warnings on CI

foonathan:
I'm currently getting a lot of warnings when building cmark.
I'm in the process of fixing those, would be nice if the project is kept clean.


nwellnhof:
Maybe we should simply compile with `/W3` instead of `/W4` under MSVC?

foonathan:
The most annoying, serious warning was the one about possible loss of data fixed in a937593293ee5a62056aa3cdbd153e8955a6a491.
And that's a potential issue, in some places where I've added casts there is a potential loss of data if the input is long enough, for example.

nwellnhof:
a937593 looks fine, but it should be noted that adding an explicit cast doesn't fix potential truncation issues. Besides, if we want this kind of warnings, then we should enable them under GCC, too (`-Wconversion`).

PR #185, opened 2017-02-05
libcmark.pc: use CMAKE_INSTALL_LIBDIR

juhp:
needed for multilib distros like Fedora

See https://bugzilla.redhat.com/show_bug.cgi?id=1416426

Particularly with `pkgconf`.


jgm:
Thanks

jgm:
@juhp How is CMAKE_INSTALL_LIBDIR supposed to be set?  It doesn't seem to be set to `lib` by default, and this has broken the default build (see #236).

juhp:
Hmm, I am not sure to be honest - I had assumed that cmake would default it to `lib`.
I am not a cmake aficionado though  - perhaps there is a better more correct way of doing this.

PR #194, opened 2017-05-05
Remove normalize as an option per #190

### LINKED ISSUE #190: Roundtripping issues with escaped entities
Currently we cannot roundtrip escaped entities through CommonMark:

``` shell
root@0063e4e160bd:/src/cmark# echo '\&trade;'
\&trade;
root@0063e4e160bd:/src/cmark# echo '\&trade;' | build/src/cmark-gfm -t commonmark
&trade;
root@0063e4e160bd:/src/cmark# echo '&trade;' | build/src/cmark-gfm -t commonmark
™
```

The issue lies in the implementation of [`render.c`'s `S_out`](https://github.com/jgm/cmark/blob/12501f1000f592ff67cca98b6733f0bfa3115f9f/src/render.c#L65): it calculates `nextc`, which is eventually passed to `commonmark.c`'s `outc`, which itself escapes `&` iff `cmark_isalpha(nextc)`.

The problem is that `S_out` is given at most one node's content at a time in `source` (and of course, often less). Without normalising text nodes, a literal (escaped) `&` is guaranteed to be in a node separate to any adjacent text nodes (as created by `handle_backslash`), and therefore this case will never trigger.

Normalizing text nodes, then, predictably fixes the problem:

``` shell
root@0063e4e160bd:/src/cmark# echo '\&trade;' | build/src/cmark-gfm -t commonmark --normalize
\&trade;
```

kivikakk:
Always normalize, fixes #190. This ensures CommonMark roundtripping works as expected.

Should we shift the option constants `CMARK_OPT_VALIDATE_UTF8` and `CMARK_OPT_SMART` forward? Unclear.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ $(ALLTESTS): $(SPEC)
 
 leakcheck: $(ALLTESTS)
 	for format in html man xml latex commonmark; do \
-	  for opts in "" "--smart" "--normalize"; do \
+	  for opts in "" "--smart"; do \
 	     echo "cmark -t $$format $$opts" ; \
 	     valgrind -q --leak-check=full --dsymutil=yes --error-exitcode=1 $(PROG) -t $$format $$opts $(ALLTESTS) >/dev/null || exit 1;\
           done; \
--- a/man/man3/cmark.3
+++ b/man/man3/cmark.3
@@ -1,4 +1,4 @@
-.TH cmark 3 "November 18, 2016" "LOCAL" "Library Functions Manual"
+.TH cmark 3 "May 05, 2017" "LOCAL" "Library Functions Manual"
 .SH
 NAME
 .PP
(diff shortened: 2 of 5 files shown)


jgm:
I don't see any compelling reason to shift the constants, and this might cause problems for wrappers that have hard-coded these.

nwellnhof:
Removing `CMARK_OPT_NORMALIZE` is an API change and breaks all code using it. I'd simply keep it around and document that it's a legacy option that has no effect.

jgm:
+++ Nick Wellnhofer [May 06 17 08:46 ]:
>   Removing CMARK_OPT_NORMALIZE is an API change and breaks all code using
>   it. I'd simply keep it around and document that it's a legacy option
>   that has no effect.

Agreed, that's a better plan.  Does someone want to submit a
PR?

kivikakk:
>  Does someone want to submit a PR?

-> #195.

PR #195, opened 2017-05-07
Revert "Remove normalize as an option per #190"

kivikakk:
Reverts jgm/cmark#194 — amending just revert the option define removal.


jgm:
Thanks!

PR #200, opened 2017-05-22
Getters for table cell alignments

kivikakk:
Fixes https://github.com/github/cmark/issues/28.

/cc @nafest


kivikakk:
I need to fix the default PR fork on this. Sorry for the noise.

PR #205, opened 2017-06-21
Don't scan past an EOL

kivikakk:
The existing negated character classes (`[^…]`) are careful to always include `\x00` in the characters excluded, but these `.` catch-alls can scan right past the terminating NUL placed at the end of the buffer by `_scan_at`.

As such, buffer overruns can occur.

### DIFF
--- a/src/scanners.c
+++ b/src/scanners.c
@@ -12622,7 +12622,7 @@ bufsize_t _scan_html_block_end_1(const unsigned char *p) {
     unsigned char yych;
     unsigned int yyaccept = 0;
     static const unsigned char yybm[] = {
-        0,  64, 64, 64, 64, 64, 64,  64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64,
+        0,  64, 64, 64, 64, 64, 64,  64, 64, 64, 0,  64, 64, 64, 64, 64, 64, 64,
         64, 64, 64, 64, 64, 64, 64,  64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64,
         64, 64, 64, 64, 64, 64, 64,  64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64,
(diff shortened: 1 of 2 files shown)


src/scanners.c
```diff
@@ -12622,7 +12622,7 @@ bufsize_t _scan_html_block_end_1(const unsigned char *p) {
     unsigned char yych;
     unsigned int yyaccept = 0;
     static const unsigned char yybm[] = {
-        64, 64, 64, 64, 64, 64, 64,  64, 64, 64, 0,  64, 64, 64, 64, 64, 64, 64,
+        0,  64, 64, 64, 64, 64, 64,  64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64,
```
kivikakk: `0,  64, 64, 64, 64, 64, 64,  64, 64, 64, 64, `

Last 64 should be 0.

kivikakk: (`.` matches any character except newline — we want to match any character except newline _and_ NUL)

jgm:
Good catch!

jgm:
@kivikakk Do you have a test case where this causes an actual problem?
The block parser grabs one line at a time, so that should be enough to guarantee that the string on which these scanners operate contains no NULL bytes (and no LF except perhaps at the end).

kivikakk:
It doesn't matter if they contain NUL bytes or not — the problem is `_scan_at` appears to be designed to explicitly add a NUL byte to the end of the buffer, presumably to cause termination of the scanning process at the end of the buffer, but `re2c` doesn't stop scanning at a NUL unless you explicitly tell it to (such as by this patch). A rule like `.*` will happily scan past the NUL.

It's really hard to make a reliable test case for because the only way to know this case is triggered is by controlling exactly where the in memory the input subject is allocated, and e.g. placing a guard page immediately after which triggers a protection error or similar. We were noticing CI failing these kinds of tests consistently, but only a very small % of the time (whenever the out of bound read would happen to cause a crash).

jgm:
Right, I understand now.

Issue #206, opened 2017-06-22
Out of bounds read with CMARK_OPT_VALIDATE_UTF8


philipturnbull:
A single-byte out of bounds read can be triggered when a markdown file contains invalid UTF8 bytes and `--validate-utf8` is enabled. This can be reproduced with:

```
$ make asan
$ echo IGNoci9zdHkKbmdsPC9zdFlZAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAFl5c3R5c3R59p6LhvWRmIo8L3N0WQpuZ2w8bmdsPC9zdDwvZw== \
| base64 -D | ./build/src/cmark --validate-utf8
=================================================================
==76495==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61400000a3c8 at pc 0x00010ed20269 bp 0x7fff50ee3610 sp 0x7fff50ee3608
READ of size 1 at 0x61400000a3c8 thread T0
    #0 0x10ed20268 in S_process_line blocks.c:1150
    #1 0x10ed1f990 in S_parser_feed blocks.c:555
    #2 0x10ed1fcca in cmark_parser_feed blocks.c:521
    #3 0x10ed6191c in main main.c:171
    #4 0x7fffbbd67234 in start (libdyld.dylib:x86_64+0x5234)

0x61400000a3c8 is located 0 bytes to the right of 392-byte region [0x61400000a240,0x61400000a3c8)
allocated by thread T0 here:
    #0 0x10eebb520 in wrap_realloc (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x56520)
    #1 0x10ed1c228 in xrealloc cmark.c:23
    #2 0x10ed558d8 in cmark_strbuf_grow buffer.c:58
    #3 0x10ed557e4 in cmark_strbuf_init buffer.c:32
    #4 0x10ed1f3e2 in cmark_parser_new_with_mem blocks.c:89
    #5 0x10ed1f54f in cmark_parser_new blocks.c:112
    #6 0x10ed617b9 in main main.c:149
    #7 0x7fffbbd67234 in start (libdyld.dylib:x86_64+0x5234)

[...]

PR #207, opened 2017-06-23
Reset bytes after (possible) UTF8 proc

kivikakk:
Fixes the issue reported in https://github.com/jgm/cmark/issues/206 by reinitializing the `bytes` local with the actual size of `parser->curline`.

/cc @jgm @philipturnbull

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -1146,6 +1146,8 @@ static void S_process_line(cmark_parser *parser, const unsigned char *buffer,
   else
     cmark_strbuf_put(&parser->curline, buffer, bytes);
 
+  bytes = parser->curline.size;
+
   // ensure line ends with a newline:
   if (bytes == 0 || !S_is_line_end_char(parser->curline.ptr[bytes - 1]))
     cmark_strbuf_putc(&parser->curline, '\n');


kivikakk:
Tested to no longer fail the sample case provided in #206.

jgm:
Thanks for the PR, @kivikakk.  I'd prefer the following patch, though:
```
diff --git a/src/blocks.c b/src/blocks.c
index 98868d3..373543e 100644
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -1147,7 +1147,8 @@ static void S_process_line(cmark_parser *parser, const unsigned char *buffer,
     cmark_strbuf_put(&parser->curline, buffer, bytes);
 
   // ensure line ends with a newline:
-  if (bytes == 0 || !S_is_line_end_char(parser->curline.ptr[bytes - 1]))
+  if (parser->curline.size == 0 ||
+      !S_is_line_end_char(parser->curline.ptr[parser->curline.size - 1]))
     cmark_strbuf_putc(&parser->curline, '\n');
 
   parser->offset = 0;
```
This makes it clearer that, once we have `parser->curline`, we no longer need `bytes`.
This should do the same -- does that look okay to you?

jgm:
OK, that's reasonable.  I'll go with your solution then.

PR #212, opened 2017-07-06
Avoid memcpy'ing NULL pointers

kivikakk:
A UBSAN warning can be triggered because the link title is an empty string:

  src/inlines.c:113:20: runtime error: null pointer passed as argument 2, which is declared to never be null

which can be triggered by:
```
[f]:_
[f]
```

The length of the memcpy is zero so the NULL pointer is not dereferenced but it
is still undefined behaviour.

(Originally github#38.)


jgm:
Thanks!

PR #213, opened 2017-07-11
Use unsigned integer when shifting

kivikakk:
(original at github#39)

A UBSAN warning can be triggered when handling a long sequence of backticks:

src/commonmark.c:98:20: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'

which can be triggered by:

```
| a | b |
| --- | --** `c```````````````````````````````- |
| c | `|d` \| e |
```


jgm:
Great!

PR #216, opened 2017-07-12
Limit nested parens in URL scanner

kivikakk:
Fixes https://github.com/jgm/cmark/issues/214.

/cc @mity @vmg


kivikakk:
Second time in a day? Okay, I'm definitely fixing this this week …

PR #217, opened 2017-07-12
> 32 nested balanced parens in a link is bananas

kivikakk:
Fixes https://github.com/jgm/cmark/issues/214 in this fork.

(This is what #216 should have been.)

/cc @mity @vmg


jgm:
Technically, this is inconsistent with the spec:

> Any number parentheses are allowed without escaping, as long as they are balanced

I suppose we should say in the spec that there's a limit?  Or at least that implementations may impose one?

kivikakk:
I think "implementations may impose a limit" is fine.

PR #219, opened 2017-07-17
Fix URL scanner

### LINKED ISSUE #193: Unescaped left angle brackets in link destination
The CommonMark spec has this to say about link destinations:
> A link destination consists of either
>
>a sequence of zero or more characters between an opening < and a closing > that contains no >spaces, line breaks, **or unescaped < or > characters**, or

Given the following input:
```
[a]

[a]: <te<st>
```
this implementation outputs:
```
<p><a href="te%3Cst">a</a></p>
```

However, the Javascript reference implementation outputs:
```
<p><a href="%3Cte%3Cst%3E">a</a></p>
```

Note the `%3C` characters surrounding `te%3Cst` are missing in the output from this implementation.  They should be present, as this should be treated as the non-`<...>` type of link definition.

This also seems to affect inline link destinations.

kivikakk:
Fixes #193.

We fallback to scanning a non `<>`-type link if we fail the `<>`-type scan. (Really need a better name for this!) Clean the interface up a little bit so the scanner only returns the actual URL, whether it includes `<>` at the start or end or not, and stop `cmark_clean_url` from trying to do that for us. (We'd end up accidentally trimming `<` and `>` from the start and end of a URL like `<a<b>` otherwise.)

/cc @jgm

### DIFF
--- a/src/inlines.c
+++ b/src/inlines.c
@@ -695,8 +695,8 @@ static cmark_node *handle_entity(subject *subj) {
   return make_str(subj->mem, cmark_chunk_buf_detach(&ent));
 }
 
-// Clean a URL: remove surrounding whitespace and surrounding <>,
-// and remove \ that escape punctuation.
+// Clean a URL: remove surrounding whitespace, and remove \ that escape
+// punctuation.
 cmark_chunk cmark_clean_url(cmark_mem *mem, cmark_chunk *url) {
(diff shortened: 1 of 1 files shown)


jgm:
Thanks.  Looks good as far as I can see!

jgm:
Having merged this, I think there's still a problem.  I'll open a new issue for it, though.

PR #221, opened 2017-08-01
Advertise --validate-utf8 in usage information

pclouds:
This option was added in 04726a7 (Added `CMARK_OPT_VALIDATE_UTF8` option. - 2015-06-16) but not "documented".

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -32,6 +32,7 @@ void print_usage() {
   printf("  --nobreaks       Render soft line breaks as spaces\n");
   printf("  --safe           Suppress raw HTML and dangerous URLs\n");
   printf("  --smart          Use smart punctuation\n");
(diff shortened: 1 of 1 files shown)


jgm:
Thanks!

pclouds:
Thanks @jgm. Quick side question about man page update though.

I see that there's a python script (and a make target) to regenerate cmark.3 from source code but not for cmark.1. Does it mean if I add a new command line option (I might, we'll see ;) I have to update cmark.1 manually?

jgm:
+++ Duy Nguyen [Aug 02 17 03:08 ]:
>   Thanks [1]@jgm. Quick side question about man page update though.
>
>   I see that there's a python script (and a make target) to regenerate
>   cmark.3 from source code but not for cmark.1. Does it mean if I add a
>   new command line option (I might, we'll see ;) I have to update cmark.1
>   manually?

Yes.

Issue #225, opened 2017-08-05
Surfacing the list bullet marker type


nathanaschbacher:
Howdy,

Currently ordered lists expose which delimiter, `.` or `)`, was used when defining a list. I'd like to add functionality which does something similar for *un*ordered lists. Such that the bullet-type is made available to the AST and/or output. So something consuming the AST or output could check for something like `plus`, `hyphen`, or `asterisk` when `+`, `-`, or `*` was used to construct the list.

Is that kind of functionality something that would be accepted upstream? I'm asking this first, because instead of adding this to cmark I could instead create an entirely separate pre-processor tool for our use case, but that would be non-ideal compared to adding support to cmark. Especially given that adding this would add balance to the current discrepancy in available AST detail between ordered and unordered lists.

If it is something that would have a chance at being accepted upstream, then there are two ways of implementing this that make sense to me.

1. Extending the existing `cmark_delim_type` in `cmark.h` to include additional delimiters for bullet-type, but that then makes this type cover more than its original intended function, which was clearly intended only for ordered lists.

2. Creating an entirely new type such as `cmark_bullet_type` in `cmark.h` to represent the different unordered list bullet types, but then this adds a whole new data type to cmark.

Either of these methods will work, but I want to make sure the solution we pursue fits in with how the cmark maintainers would prefer the extra context/metadata be captured and represented.

Any guidance on any of the above would be very helpful.

Thank you.

nathanaschbacher:
Having now partially implemented both options 1 and 2 above I have some thoughts about their respective impacts.

Option 1: Fits in with the cmark project structure and architecture of the parser and various output format generators as they exist today.

Option 2: Seems more logically intuitive in the abstract, but is a bit more baroque when actually trying to integrate the change in cmark.

As a result I'm currently more inclined to go for Option 1 unless otherwise influenced.

nathanaschbacher:
I've nearly got a PR ready for Option 1. All tests pass and I added an additional "api" test to the suite. I've also updated the racket-lang wrapper and validated that when doing `-t commonmark` the bullet character is preserved.

The only oddity is the `CMARK_NO_DELIM` symbol that was already present in `cmark_delim_type`. It's now not necessary, but was previously being used as a kind of *default value* to signal a bulletted list in some scenarios.

Probably the proper way to handle this would be to remove `CMARK_NO_DELIM` from `cmark_delim_type` entirely, and locate all the places which check against it or return it and instead check against the list type or return a more specific status, respectively. 

[...]

PR #226, opened 2017-08-06
Add exposing list markers *, -, +, or numeric to AST and API.

nathanaschbacher:
Prior to this commit the node struct for a list contained a member to
store whatever char was parsed from the markdown, but this value
wasn't exposed through the API, nor was it made available in the
output of the AST in any format.

Following this commit a `cmark_marker_type` struct has been added to
capture various forms of list markers: `CMARK_NUMERIC_MARKER` (for
ordered lists) and `CMARK_ASTERISK_MARKER`, `CMARK_HYPHEN_MARKER`, &
`CMARK_PLUS_MARKER` (for unordered lists). The value of which is
stored in the `marker` member of the `cmark_list` node struct.

This enables two primary features:

1. The bullet marker types are now preserved when converting from
markdown to markdown.
2. The XML representation of the AST now has a `marker="hyphen"` (as
well as numeric, asterisk, and plus) as an attribute on the list
nodes. This is similar to the `delim="period"` attribute available on
ordered list nodes.

Tests have been added and are passing. The original test suite still
passes as well. This commit also updates the racket-lang wrapper to
support these new features as well as fixes a minor bug for a missing
`no_delim` symbol because `NO_DELIM` wasn't defined in cmark.h

If you'd like to see examples of the different results of running this through the updated `cmark` executable, then take a look at this Gist: https://gist.github.com/nathanaschbacher/93f45e938aef671a07e8c9f8850115cf

Finally, this should close issue #225 that I opened earlier today.

Cheers!

### DIFF
--- a/api_test/main.c
+++ b/api_test/main.c
@@ -103,6 +103,8 @@ static void accessors(test_batch_runner *runner) {
   cmark_node *bullet_list = cmark_node_next(heading);
   INT_EQ(runner, cmark_node_get_list_type(bullet_list), CMARK_BULLET_LIST,
          "get_list_type bullet");
+  INT_EQ(runner, cmark_node_get_list_marker(bullet_list), CMARK_ASTERISK_MARKER,
+         "get_list_marker asterisk");
   INT_EQ(runner, cmark_node_get_list_tight(bullet_list), 1,
          "get_list_tight tight");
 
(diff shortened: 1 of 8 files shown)


nathanaschbacher:
I also have a similar branch that doesn't create the concept of a "marker" and instead just overloads the "deliminator" concept to include `asterisk`, `hyphen`, and `plus` if that's preferable.

Likewise for a version that doesn't bother to convert to a term and just keeps the syntax, like `marker="*"` instead of `marker="asterisk"`, but that seemed to break with the convention used by `cmark` already.

gjtorikian:
Could there be a review/consideration for this?

PR #228, opened 2017-08-09
Add sourcepos info for inlines

kivikakk:
In github#53 we added sourcepos information for inlines; here's that PR targeting upstream, if you'd like!

/cc @jgm


kivikakk:
This produces incorrect values in some cases. Sample:

```
kahekkass ~/github/cmark master$ echo '**Hello**' | build/src/cmark -t xml --sourcepos
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document sourcepos="1:1-1:9" xmlns="http://commonmark.org/xml/1.0">
  <paragraph sourcepos="1:1-1:9">
    <strong sourcepos="1:1-1:9">
      <text sourcepos="1:3-1:7">Hello</text>
    </strong>
  </paragraph>
</document>
kahekkass ~/github/cmark master$ echo '# **Hello**' | build/src/cmark -t xml --sourcepos
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document sourcepos="1:1-1:11" xmlns="http://commonmark.org/xml/1.0">
  <heading sourcepos="1:1-1:11" level="1">
    <strong sourcepos="1:1-1:9">
      <text sourcepos="1:3-1:7">Hello</text>
    </strong>
  </heading>
</document>
```

In the second case, the `<strong …>` should have a different sourcepos to the same element in the first case (because it's two characters over), but it doesn't. Will update this PR soon.

kivikakk:
Fix made; there were two issues:

1. We weren't adding an offset to the columns of inlines nested inside heading elements.

2. Newlines in non-text inlines (code, raw HTML) were not being noticed, and we weren't updating column/line numbers accordingly.

kivikakk:
Resolved conflicts from latest master to stop this falling too far behind; @jgm, any interest in merging? If not, I'll close and stop keeping it current. :+1:

nwellnhof:
The new field `internal_offset` in `struct cmark_node` seems to be used only for headings. I think we could get rid of that field if we added an `offset` parameter to `cmark_parse_inline` that contains the current offset from `struct cmark_parser`.

kivikakk:
I've taken a look into this and I'm not sure I see an obvious way to propagate this information correctly without storing it in the node.

jgm:
@kivikakk I am interested, sorry for the silence on this, and thanks for updating the PR.

kivikakk:
@jgm not at all, thank you!

PR #243, opened 2017-11-05
install cmake exports to lib/cmark

kivikakk:
`lib/cmark/…`, not `lib/cmake/…`. Project names are too similar!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -148,7 +148,7 @@ if(CMARK_SHARED OR CMARK_STATIC)
     DESTINATION include
     )
 
-  install(EXPORT cmark DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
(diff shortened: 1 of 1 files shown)


jgm:
I had always assumed that `lib/cmake` was right for these files
(`cmark-release.cmake` and `cmark.cmake`), because these are
used by cmake itself to locate libraries.

But I'm no cmake expert.  Why don't you comment on commit
9fe857d827b1fdb585e12d8387c3cb5fb5392c61
which originally added this, and see if the author
clarifies?

kivikakk:
Oh, you're very possibly right. Hrm. I'll do so, thank you.

kivikakk:
I've rereviewed docs, and this is in fact correct. :+1:

jgm:
https://cmake.org/cmake/help/v2.8.8/cmake.html#command:find_package

PR #244, opened 2017-11-06
Fix header install

### LINKED ISSUE #62: freeing cmark_render_* buffers
I am writing bindings for C# and I saw that there are 4 functions which take a buffer immediately and do some rendering:

```c
/** Render a 'node' tree as XML.
 */
CMARK_EXPORT
char *cmark_render_xml(cmark_node *root, int options);

/** Render a 'node' tree as an HTML fragment.  It is up to the user
 * to add an appropriate header and footer.
 */
CMARK_EXPORT
char *cmark_render_html(cmark_node *root, int options);

/** Render a 'node' tree as a groff man page, without the header.
 */
CMARK_EXPORT
char *cmark_render_man(cmark_node *root, int options);

/** Render a 'node' tree as a commonmark document.
 */
CMARK_EXPORT
char *cmark_render_commonmark(cmark_node *root, int options, int width);
```

However, I have no access to free from C# (Well one can code it but it is considered to be not good).

Maybe we could expose the free function (cmark_render_free?) so a user who uses ffi bindings could just tap into that?

kivikakk:
Fixes #62.

/cc @felixonmars


kivikakk:
One day I will fix this. :/

Issue #247, opened 2017-11-12
Building with CMARK_SHARED=OFF fails


ericpruitt:
I'm attempting to build a static cmark binary by setting the cmake option "CMARK_SHARED" to "OFF", but the cmark binary fails to build. This is the script I'm using:

```bash
rm -rf CMakeFiles/ CMakeCache.txt build/ &&
(
    mkdir -p build &&
    cd build/ &&
    cmake -G "Unix Makefiles" \
        -DCMAKE_BUILD_TYPE= \
        -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
        -DCMARK_TESTS=OFF -DCMARK_SHARED=OFF .. &&
    make
)
```

The build fails because the compiler is unable to locate some source files. All failed builds show the exact same error:

```C
In file included from /home/ericpruitt/projects/mdlint/cmark/src/node.h:11:0,
                 from /home/ericpruitt/projects/mdlint/cmark/src/node.c:5:
/home/ericpruitt/projects/mdlint/cmark/src/cmark.h:5:26: fatal error: cmark_export.h: No such file or directory
 #include <cmark_export.h>
                          ^
compilation terminated.
```

Here's the diff of the generated Makefiles with CMAKE_SHARED=ON/OFF:

```diff
--- CMAKE_SHARED=ON      2017-11-11 19:01:31.807657734 -0800
+++ CMAKE_SHARED=OFF      2017-11-11 19:08:02.069241357 -0800
@@ -168,19 +168,6 @@
 .PHONY : cmark/fast

 #=============================================================================
-# Target rules for targets named libcmark
-
-# Build rule for target.
-libcmark: cmake_check_build_system
-       $(MAKE) -f CMakeFiles/Makefile2 libcmark
-.PHONY : libcmark
-
-# fast build rule for target.
-libcmark/fast:
-       $(MAKE) -f src/CMakeFiles/libcmark.dir/build.make src/CMakeFiles/libcmark.dir/build
-.PHONY : libcmark/fast
-
-#=============================================================================
 # Target rules for targets named libcmark_static

 # Build rule for target.
@@ -206,7 +193,6 @@
        @echo "... list_install_components"
        @echo "... install"
        @echo "... cmark"
-       @echo "... libcmark"
        @echo "... libcmark_static"
 .PHONY : help
```

A year and a half ago, I opened #129 , and I was able to build just a static cmark binary by modifying CMakeFiles.txt at your recommendation, but that no longer works either:

```bash
[...]

PR #248, opened 2017-11-14
Add static export header & deny building both shared and static at once

kivikakk:
Fixes https://github.com/commonmark/cmark/issues/247 by adding an export header for the static library, and denying building both shared and static (due to issues raised in https://github.com/commonmark/cmark/issues/247#issuecomment-344142209).

/cc @nwellnhof


nwellnhof:
See #249 for a better fix that still allows building both shared and static libraries.

PR #254, opened 2018-02-21
Check for empty buffer when rendering

philipturnbull:
This minor bug came in via GitHub's Bug Bounty program. I did some digging and we didn't find this bug via fuzzing because we discard all empty documents in `test/cmark-fuzz.c`. I tweaked the fuzzing harness to accept empty documents.

I also noticed we use a fixed width when fuzzing. Allow the `width` parameter to be fuzzed too for better coverage.


jgm:
Belated thanks!

PR #257, opened 2018-07-05
Add xml:space="preserve" in XML output when appropriate

pclouds:
With current HTML escaping, sometimes we may produce an XML tag like

    <code_block>  </code_block>

Many XML parsers consider these spaces insignificant and strip them out but we need this. There's actually a test case like this in spec.txt (search "A code block can have all empty lines as its content:") and without proper hinting, an external xml->html converter will fail the spec.

XML standard covers this case. If xml:space is "preserve", then whitespaces in an all-space node are significant and should be kept (this attribute does not affect tags that have whitespaces mixed with
non-whitespaces).

Add this hint for text, code, code_block, html_inline and html_block tags. html_inline and html_block are probably not so problematic but I'm just hesistant to break them out as a separate "case" block.


pclouds:
[1] suggests it only affects the "all spaces" case. At least with the SSAX parser i'm using, whenever there's non-whitespace character, the whole tag content is preserved. [2] has examples that preserve consecutive whitespaces in the middle of the tag though.

So I guess the simplest way is always add xml:space="preserve" for these tags. Or we could detect at least one whitespace and only add xml:space="preserve" then. The former sounds a lot simpler. Which one do you prefer?

[1] http://www.xmlplease.com/xml/xmlspace/#s3.
[2] https://msdn.microsoft.com/en-us/library/ms256097(v=vs.110).aspx

pclouds:
Version 2 unconditionally adds xml:space="preserve" too all text, code, code_block and html_* tags. We could tweak it further to not add it unless necessary but this is definitely safer than version 1.

pclouds:
@jgm Is there anything I can do to move this forward?

pclouds:
ping...

jgm:
Thanks.  Sorry for the delay!

PR #265, opened 2018-09-05
build: only attempt to install MSVC system libraries on Windows

compnerd:
Newer versions of CMake attempt to query the system for information about the VS
2017 installation.  Unfortunately, this query fails on non-Windows systems when
cross-compiling:

	cmake_host_system_information does not recognize <key> VS_15_DIR

CMake will not find these system libraries on non-Windows hosts anyways, and we
were silencing the warnings, so simply omit the installation when
cross-compiling to Windows.


jgm:
Thanks

compnerd:
My pleasure!

PR #269, opened 2018-09-21
Fix cmake warning about CMP0048, again

foonathan:
I fixed that warning in #137, by setting the policy to OLD, but now it is being deprecated.

The correct fix is to set it to NEW conditionally and use the `project(.... VERSION)` command which sets all those variables. The old code is kept for compatibility with cmake 2.x.

Given that CMake 3.0 was released in 2014, maybe it is time to set the minimum version to 3.0? This would clean that up.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -59,7 +59,6 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmark_version.h.in
 include (GenerateExportHeader)
 
 add_executable(${PROGRAM} ${PROGRAM_SOURCES})
-add_compiler_export_flags()
 
 # Disable the PUBLIC declarations when compiling the executable:
 set_target_properties(${PROGRAM} PROPERTIES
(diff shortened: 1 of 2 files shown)


jgm:
Yes, even the previous debian stable (jessie) has cmark >= 3, so maybe it would make sense just to require that.  @nwellnhof do you have thoughts on this?

nwellnhof:
I don't have a strong opinion.

jgm:
Revisiting this:  I am fine with requiring cmake >= 3.0.
@foonathan do you want to revise this PR, simplifying it and requiring cmark 3?

foonathan:
Yes, will do.

src/CMakeLists.txt
```diff
@@ -59,7 +59,6 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/cmark_version.h.in
 include (GenerateExportHeader)
 
 add_executable(${PROGRAM} ${PROGRAM_SOURCES})
-add_compiler_export_flags()
```
jgm: Can you explain why this line is removed?

foonathan: The function is deprecated see commit message.

jgm:
Great, thanks!

PR #286, opened 2019-03-14
Add automatic configuration of compiler to get large file support

kivikakk:
Upstreaming @keith-packard's large file support PR from https://github.com/github/cmark-gfm/pull/138.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ endif()
 project(cmark)
 
 include("FindAsan.cmake")
+include("CheckFileOffsetBits.cmake")
 
 if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     message(FATAL_ERROR "Do not build in-source.\nPlease remove CMakeCache.txt and the CMakeFiles/ directory.\nThen: mkdir build ; cd build ; cmake .. ; make")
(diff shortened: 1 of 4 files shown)


jgm:
Can you say more about this @keith-packard?  What sort of actual case does this fix?

@mity's [comment](https://github.com/github/cmark-gfm/pull/138#issuecomment-473714605) on the linked issue is correct:  cmark needs to read the entire file into memory to parse it, so it's hard to see how reading > 4 GB files on a 32-bit system would work anyway.

keith-packard:
> Can you say more about this @keith-packard? What sort of actual case does this fix?
> 
> @mity's [comment](https://github.com/github/cmark-gfm/pull/138#issuecomment-473714605) on the linked issue is correct: cmark needs to read the entire file into memory to parse it, so it's hard to see how reading > 4 GB files on a 32-bit system would work anyway.

On 32-bit systems, using '#define _FILE_OFFSET_BITS 64' changes the ABI so that various types are made 64-bits instead of 32-bits. It's a (largely) source-compatible change, but not binary-compatible, so applications have to opt-in to the switch. The result is a program which can support files > 4GB on 32-bit systems.

This issue was raised by debian's automatic package testing on the 32-bit version of the package; it doesn't affect the 64-bit version of the package at all.

kivikakk:
This issue has become stale, so closing.

Issue #296, opened 2019-04-24
Source position seems off


chriseidhof:
Give the following input:

```
X
    **p**
```

The cmark library reports (through `cmark_node_get_start_column`) that the strong node starts at line 2, column 1, and ends at column 5. I would have expected it to start at column 5. Is this a bug, or is my interpretation of start column wrong?

honghaoz:
Looks like this is related to the soft breaks:

<img width="281" alt="image" src="https://user-images.githubusercontent.com/3926785/56697308-1e908180-66a3-11e9-8960-c77bcdecf530.png">

jgm:
I'll refer this to @kivikakk who added inline source positions in #228.

chriseidhof:
Thank you. I added a passing test case in #297, hopefully that makes it easier to get started (or if this is expected behavior, we can merge it as-is).

honghaoz:
For more information, I found that for md like
```
A
  B
```
For the 1st pass when generating the blocks. The container block's string content already skips the spaces. So the container block node has text "A\nB\n" not "A\n  B\n", that's probably why the source position is off for the [lazy continuation lines](https://spec.commonmark.org/0.29/#lazy-continuation-line).

honghaoz:
Related issue: https://github.com/commonmark/cmark/issues/204

kivikakk:
This issue becomes more pronouncedly odd when you try the inverse:

```md
  A
B
```

produces

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document sourcepos="1:1-2:1" xmlns="http://commonmark.org/xml/1.0">
  <paragraph sourcepos="1:3-2:1">
    <text sourcepos="1:3-1:3" xml:space="preserve">A</text>
    <softbreak />
    <text sourcepos="2:3-2:3" xml:space="preserve">B</text>
  </paragraph>
</document>
```

The lazy continuation means we end up treating everything as though it started at column 3.

honghaoz:
@kivikakk I think that's because the first line has two leading spaces. If you try to change it to 3 or 1, you will see the start column of following lines changed.

I briefly went through the parsing logic:
https://github.com/commonmark/cmark/blob/master/src/blocks.c#L706 this line advances the `parser->offset` and `parser->column` for the 2nd line.

https://github.com/commonmark/cmark/blob/master/src/blocks.c#L186-L187 this line only adds the skipped text to the container node.

[...]

PR #297, opened 2019-04-25
[wip] Add test info for softbreak

chriseidhof:
This test actually passes. I would have expected it to have a different source range for the `strong` and `text` nodes, as per #296.


honghaoz:
I believe this worth a test case as well.
```
  This is
    **strong**
```
Notice the leading spaces in the 1st line.

In addition,
```
1. This is
  **Strong**
```
```
1. This is
    **Strong**
```
```
1.   This is
    **Strong**
```
(Notice leading spaces in the 2nd line, seems it's related to the padding (3 spaces) of the parent list block)
These are some test cases I found useful to add.

PR #300, opened 2019-05-25
Remove "-rdynamic" flag for static builds

ericpruitt:
I ran into problems trying to build the statically linked cmark executable using [musl libc](https://www.musl-libc.org/) that was caused by the "-rdynamic" flag being implicitly added to the build command. The resulting executable would had references to a musl libc shared object instead of being hermetic as expected. I'm not sure if this has any unintended consequences, but I at least have no problems building a dynamically linked executable with this patch applied on my Linux and macOS machines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ option(CMARK_STATIC "Build static libcmark library" ON)
 option(CMARK_SHARED "Build shared libcmark library" ON)
 option(CMARK_LIB_FUZZER "Build libFuzzer fuzzing harness" OFF)
 
-if(NOT APPLE AND CMARK_STATIC)
+if(LINUX AND STATIC)
   string(REGEX REPLACE "( |^)-rdynamic( |$)" " " CMAKE_SHARED_LIBRARY_LINK_C_FLAGS ${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,10 @@ option(CMARK_STATIC "Build static libcmark library" ON)
 option(CMARK_SHARED "Build shared libcmark library" ON)
 option(CMARK_LIB_FUZZER "Build libFuzzer fuzzing harness" OFF)
 
-if(CMARK_STATIC)
+# The Linux modules distributed with CMake add "-rdynamic" to the build flags
+# which is incompatible with static linking under certain configurations.
+# Unsetting CMAKE_SHARED_LIBRARY_LINK_C_FLAGS ensures this does not happen.
+if(CMARK_STATIC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
   SET(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ option(CMARK_STATIC "Build static libcmark library" ON)
 option(CMARK_SHARED "Build shared libcmark library" ON)
 option(CMARK_LIB_FUZZER "Build libFuzzer fuzzing harness" OFF)
 
-if(LINUX AND CMARK_STATIC)
+if(CMARK_STATIC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
   string(REGEX REPLACE "( |^)-rdynamic( |$)" " " CMAKE_SHARED_LIBRARY_LINK_C_FLAGS ${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ option(CMARK_SHARED "Build shared libcmark library" ON)
 option(CMARK_LIB_FUZZER "Build libFuzzer fuzzing harness" OFF)
 
 if(CMARK_STATIC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
-  string(REGEX REPLACE "( |^)-rdynamic( |$)" " " CMAKE_SHARED_LIBRARY_LINK_C_FLAGS ${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS})
+  SET(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
 endif()
 
 add_subdirectory(src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ option(CMARK_STATIC "Build static libcmark library" ON)
 option(CMARK_SHARED "Build shared libcmark library" ON)
 option(CMARK_LIB_FUZZER "Build libFuzzer fuzzing harness" OFF)
 
-if(LINUX AND STATIC)
+if(LINUX AND CMARK_STATIC)
   string(REGEX REPLACE "( |^)-rdynamic( |$)" " " CMAKE_SHARED_LIBRARY_LINK_C_FLAGS ${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS})
 endif()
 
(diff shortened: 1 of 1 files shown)


ericpruitt:
The CI pipeline shows this fails to build with MSVC due to the number of arguments used with the REGEX command. I have no experience with C development on Windows, and I'm a CMake novice, so I'm not sure how I should address the build failure.

jgm:
I'm not a cmake expert either, but I'm sure this isn't the right approach to fix this problem.  (If there is a problem: it would be good to have a fuller account of the problem, so we can confirm it.)  Probably if something needs to be changed, it's in `src/CMakeLists.txt`, where the parameters for static and dynamic builds are being defined.

ericpruitt:
> Probably if something needs to be changed, it's in `src/CMakeLists.txt`, where the parameters for static and dynamic builds are being defined.

I don't know if this makes a difference, but from what I saw when using CMake with verbosity/debugging, "-rdynamic" is added in based on rules in my systems global CMake configuration files, not something in the cmark build rules.

> (If there is a problem: it would be good to have a fuller account of the problem, so we can confirm it.)

In the text below, I've gone through the process of generating the faulty build by generating the musl compiler wrapper (musl-gcc) and using the wrapper to compile cmark:

**Extract musl and generate musl-gcc**
```
$ tar xf musl-1.1.20.tar.gz
$ cd musl-1.1.20/
musl-1.1.20$ ./configure --disable-shared --prefix=$PWD
checking for C compiler... gcc
checking whether C compiler works... yes
...
checking preprocessor condition __ILP32__... false
checking whether compiler's long double definition matches float.h... yes
checking preprocessor condition __FAST_MATH__... false
creating config.mak... done
musl-1.1.20$ make install
musl-1.1.20$ make install
mkdir -p lib
mkdir -p obj
...
nistd/unlinkat.lo obj/src/unistd/usleep.lo obj/src/unistd/write.lo obj/src/unistd/writev.lo
ranlib lib/libc.a
./tools/install.sh -D -m 644 lib/libc.a ...musl-1.1.20/lib/libc.a
musl-1.1.20$ ls bin/
musl-gcc
```

**Prepare the cmark build**
```
musl-1.1.20$ cd ../cmark-src/
cmark-src$ mkdir build && cd build
build$ CFLAGS='-static -fpie -fPIC' CFLAGS='-O3 -static -fpie -fPIC' CC=.../musl-1.1.20/bin/musl-gcc cmake -G "Unix Makefiles" -DCMARK_TESTS=OFF -DCMARK_SHARED=OFF ..
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is GNU 6.3.0
-- Check for working C compiler: .../musl-1.1.20/bin/musl-gcc
-- Check for working C compiler: .../musl-1.1.20/bin/musl-gcc -- works
...
-- Generating done
-- Build files have been written to: .../cmark/build
```

**Create the executable**
```
build$ make
Scanning dependencies of target libcmark_static
[  4%] Building C object src/CMakeFiles/libcmark_static.dir/node.c.o
[  9%] Building C object src/CMakeFiles/libcmark_static.dir/buffer.c.o
...
[ 95%] Building C object src/CMakeFiles/cmark.dir/main.c.o
[100%] Linking C executable cmark
[100%] Built target cmark
```

**Test the executable**
```
build$ ./src/cmark --help
bash: ./src/cmark: No such file or directory
(127)
```

**This shows the embedded shared object path which shouldn't be there.**
```
build$ strings src/cmark | head -n5
/lib/ld-musl-x86_64.so.1
__stdout_used
stdout
__lctrans_cur
__strtoull_internal
```

jgm:
@nwellnhof do you have any insight on this issue?

nwellnhof:
Seems like a CMake issue. The option probably comes from here: https://github.com/Kitware/CMake/blob/07cfb18f9d29cfc0588ede928846a03ec5599c48/Modules/Platform/Linux-GNU.cmake

Maybe setting `-DCMAKE_SYSTEM_NAME=Generic` helps.

ericpruitt:
> Maybe setting `-DCMAKE_SYSTEM_NAME=Generic` helps.

That doesn't make a difference for me:

```
build$ CFLAGS='-static -fpie -fPIC' CFLAGS='-O3 -static -fpie -fPIC' CC=.../musl-1.1.20/bin/musl-gcc cmake -G "Unix Makefiles" -DCMARK_TESTS=OFF -DCMARK_SHARED=OFF -DCMAKE_SYSTEM_NAME=Generic ..
-- The C compiler identification is GNU 6.3.0
-- The CXX compiler identification is Clang 3.8.1
-- Check for working C compiler: /home/ericpruitt/emus/core/musl-1.1.20/bin/musl-gcc
-- Check for working C compiler: /home/ericpruitt/emus/core/musl-1.1.20/bin/musl-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/clang++
-- Check for working CXX compiler: /usr/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test HAVE_FLAG_ADDRESS_SANITIZER
-- Performing Test HAVE_FLAG_ADDRESS_SANITIZER - Failed
-- Performing Test HAVE_FLAG_SANITIZE_ADDRESS
-- Performing Test HAVE_FLAG_SANITIZE_ADDRESS - Failed
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY - Success
-- Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY
-- Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY - Success
-- Performing Test COMPILER_HAS_DEPRECATED_ATTR
-- Performing Test COMPILER_HAS_DEPRECATED_ATTR - Success
-- Looking for stdbool.h
-- Looking for stdbool.h - found
-- Performing Test HAVE___BUILTIN_EXPECT
-- Performing Test HAVE___BUILTIN_EXPECT - Success
-- Performing Test HAVE___ATTRIBUTE__
-- Performing Test HAVE___ATTRIBUTE__ - Success
-- Configuring done
-- Generating done
-- Build files have been written to: /home/ericpruitt/emus/core/cmark/build
build$ ./src/cmark --help
bash: ./src/cmark: No such file or directory
(127)
```

ericpruitt:
The only place I see mention of "rdynamic" in the CMake source is in the modules for Linux, so I've updated the conditional in my patch accordingly, and the build now succeeds on MSVC.

nwellnhof:
I think you have to put the final `${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS}` in double quotes to avoid an error if the variable isn't defined.

jgm:
I found [this post](https://cmake.org/pipermail/cmake/2009-October/032948.html) which recommends simply
```
SET(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
```
That should be safe for a static build, right?  And then we don't need to worry about regexes.

ericpruitt:
> `SET(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)`

That works. I've updated pull request.

jgm:
Should we also now remove the check for Linux?  If I recall, it was only necessary because of the regex stuff.

ericpruitt:
I would argue that it should be scoped to Linux to minimize the chances of collateral damage because that's the only platform where this problem is known to exist since, when I grepped for "rdynamic" in the CMake repositories, the only matches were in the Linux modules directory. However, I'll do whatever's requested to get this merged as long as it fixes the problem for me `:P`, so I've updated the pull request accordingly.

jgm:
Your argument is reasonable.  I don't feel strongly about it either way.  If you want to revert the last change, that's fine, I'll merge it either way!

ericpruitt:
>  If you want to revert the last change, that's fine, I'll merge it either way!

Thanks, done.

jgm:
Thanks!

PR #301, opened 2019-06-01
Clarify which backtick strings

johelegp:
A code span can have more than its two surrounding ones, like ` `` `.

### DIFF
--- a/test/spec.txt
+++ b/test/spec.txt
@@ -5854,7 +5854,7 @@ preceded nor followed by a backtick.
 
 A [code span](@) begins with a backtick string and ends with
 a backtick string of equal length.  The contents of the code span are
-the characters between the two backtick strings, normalized in the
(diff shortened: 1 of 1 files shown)


jgm:
I didn't think the original was unclear, but if you found this clearer it's fine with me.

mity:
Shouldn't this primarily go into commonmarks-spec repo? This way the change gets lost when you update the specification to next version.

jgm:
Yes, I missed that completely.  @johelegp - this test file is simply copied from commonmark-spec when it changes, so changes here will be overridden.  commonmark-spec is the place to submit a PR for spec.txt.

johelegp:
Thank you. I'll make sure to open this and future PRs there.

waldyrious:
Cross-referencing the PR in the spec repo, for convenience: https://github.com/commonmark/commonmark-spec/pull/589

PR #319, opened 2019-12-22
build: cleanup CMake 

compnerd:
Some minor cleanups that I noticed which would improve things.  This allows building on platforms which do not use libstdc++.  There are still further refinements which can be made, but, this improves the state of things.


jgm:
Many thanks.  I barely know what I'm doing with cmake.
@nwellnhof do you have any objections to any of these changes?

nwellnhof:
Looks good. I'm only curious about the claim that `/TP` is forced on all MSVC builds. All the tests pass, so this seems to be true. But why? Then we might also consider adding `/TC` when building with newer MSVC versions, although it's not strictly necessary.

compnerd:
@jgm my pleasure :)

@nwellnhof, yeah, that is very subtle, and an accident that I caused.  In fbe364a08becb758cebe91d63ecdfe634c1601a7 I hoisted the shared configuration between the `api_test` and `src` directory.  The variant that I used was from the `api_test` directory which was subtly different: it always set `/TP` unlike the `src` version which limited it to the older releases only. If that is truly a difference that should be preserved, Im happy to adjust the stack to do so.  However, after that commit, `/TP` is passed whenever using MSVC.  At that point, setting it for the older release becomes redundant (though CMake should unique the flags).  I figure that less build configuration is better.

jgm:
Whoops, this PR caused problems with our automatic fuzzing setup:  see the build log at

https://oss-fuzz-build-logs.storage.googleapis.com/log-b8018641-3053-4be0-80c1-f17c8d1ed4e8.txt

```
Step #4: -- Performing Test HAVE_FLAG_SANITIZE_ADDRESS - Success
Step #4: CMake Error at CMakeLists.txt:46 (add_compile_definitions):
Step #4:   Unknown CMake command "add_compile_definitions".
Step #4: 
Step #4: 
Step #4: -- Configuring incomplete, errors occurred!
```

Can you fix?

compnerd:
Ugh, `add_compile_definitions` apparently was introduced in 3.12.  I'll move that to `add_compile_options` as a workaround.

nwellnhof:
> However, after that commit, /TP is passed whenever using MSVC. At that point, setting it for the older release becomes redundant (though CMake should unique the flags). I figure that less build configuration is better.

Then I'd prefer to keep the version check which also serves as implicit documentation why the flag is needed.

waldyrious:
> I'd prefer to keep the version check which also serves as implicit documentation why the flag is needed.

Even better if a comment is added as _explicit_ documentation.

PR #321, opened 2019-12-24
build: replace `add_compile_definitions`

compnerd:
Replace `add_compile_definitions` with `add_compile_options` since the
former was introduced in 3.12.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ if(MSVC)
   add_compile_options($<$<COMPILE_LANGUAGE:C>:/W4>)
   add_compile_options($<$<COMPILE_LANGUAGE:C>:/wd4706>)
   add_compile_options($<$<COMPILE_LANGUAGE:C>:/TP>)
-  add_compile_definitions($<$<COMPILE_LANGUAGE:C>:/D_CRT_SECURE_NO_WARNINGS>)
+  add_compile_options($<$<COMPILE_LANGUAGE:C>:/D_CRT_SECURE_NO_WARNINGS>)
 elseif(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES Clang)
   add_compile_options($<$<COMPILE_LANGUAGE:C>:-Wall>)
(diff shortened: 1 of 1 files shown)


compnerd:
@jgm sorry about the compatibility issue, this should fix that.

jgm:
Thank you!

jgm:
@compnerd I've been seeing this build failure on our fuzzing setup; I don't know if it has to do with your changes, but they're the most recent changes to the build infrastructure.  Also paging @nwellnhof 

https://oss-fuzz-build-logs.storage.googleapis.com/log-0d4dbbcd-58ba-4341-9359-708e4381d3df.txt

The error is
```
Step #4: clang-10: error: no such file or directory: 'build/src/libcmark.a'
```
Is it possible that one of the changes you made broke this?
Here is the bug report: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19805#c2

compnerd:
I think its probably related ...

```
Step #4: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -fsanitize=fuzzer cmark_fuzzer.o build/src/libcmark.a -o /workspace/out/address/cmark_fuzzer
```

Note that it is trying to link against `libcmark.a`, yet the build will be producing `libcmark_static.a` as per https://github.com/commonmark/cmark/blob/master/src/CMakeLists.txt#L6 (which existed prior to my changes).

I don't immediately see where this command is coming from:

```
Step #4: + clang++ -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++ -fsanitize=fuzzer cmark_fuzzer.o build/src/libcmark.a -o /workspace/out/address/cmark_fuzzer
```

nwellnhof:
No, it was `libcmark.a` before. This should fix it: https://github.com/commonmark/cmark/pull/324

PR #322, opened 2020-01-04
fix -Wconst-qual warning

compnerd:
The string literal being assigned is const, but the assignment looses
the constness of this string.  This enables building with `/Zc:strictString`
with MSVC as well.


jgm:
Thanks!

PR #324, opened 2020-01-12
Fix CMake generator expression checking for MSVC

nwellnhof:


jgm:
Thank you! cmake syntax still a total mystery to me.

PR #326, opened 2020-01-19
Rework node struct

### LINKED ISSUE #309: Inline nodes can reference text data of parent block
When parsing inline nodes, some pieces of text are kept in the parent block's content buffer and referenced from inline children as non-allocated `cmark_chunk` pointing directly into the parent's buffer. If the parent is freed, these pointers become invalid. This can lead to memory corruption. for example when moving inline nodes to another tree and deleting the old parent.

I'd suggest to copy all text data to inline children (i. e. use `chunk_clone` instead of `chunk_dup`). Then we could also think about removing the content field for block nodes.

nwellnhof:
Fixes #309.

### DIFF
--- a/src/node.h
+++ b/src/node.h
@@ -37,8 +37,8 @@ typedef struct {
 } cmark_heading;
 
 typedef struct {
-  cmark_chunk url;
-  cmark_chunk title;
+  unsigned char *url;
+  unsigned char *title;
 } cmark_link;
 
 typedef struct {
--- a/src/node.h
+++ b/src/node.h
@@ -8,13 +8,9 @@ extern "C" {
 #include <stdio.h>
 #include <stdint.h>
 
+#include "config.h"
 #include "cmark.h"
 #include "buffer.h"
-
-typedef struct {
-  unsigned char *data;
-  bufsize_t len;
-} cmark_literal;
 
 typedef struct {
   int marker_offset;
@@ -28,7 +24,6 @@ typedef struct {
 
 typedef struct {
   unsigned char *info;
-  unsigned char *literal;
   uint8_t fence_length;
   uint8_t fence_offset;
   unsigned char fence_char;
@@ -57,7 +52,7 @@ enum cmark_node__internal_flags {
 };
 
 struct cmark_node {
-  cmark_strbuf content;
+  cmark_mem *mem;
 
   struct cmark_node *next;
   struct cmark_node *prev;
@@ -66,6 +61,9 @@ struct cmark_node {
   struct cmark_node *last_child;
 
   void *user_data;
+
+  unsigned char *data;
+  bufsize_t len;
 
   int start_line;
   int start_column;
@@ -76,7 +74,6 @@ struct cmark_node {
   uint16_t flags;
 
   union {
-    cmark_literal literal;
     cmark_list list;
     cmark_code code;
     cmark_heading heading;
@@ -86,9 +83,6 @@ struct cmark_node {
   } as;
 };
 
-static CMARK_INLINE cmark_mem *cmark_node_mem(cmark_node *node) {
-  return node->content.mem;
-}
 CMARK_EXPORT int cmark_node_check(cmark_node *node, FILE *out);
 
 #ifdef __cplusplus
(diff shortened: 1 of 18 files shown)


src/node.h
```diff
@@ -17,11 +17,11 @@ typedef struct {
 } cmark_literal;
 
 typedef struct {
-  cmark_list_type list_type;
   int marker_offset;
   int padding;
   int start;
-  cmark_delim_type delimiter;
+  unsigned char list_type;
```
jgm: I'm just curious (as a C ignoramus) why this is needed.  Does the compiler default to using more than one byte for a cmark_delim_type?

jgm:
This looks good to me!

Query: does anything in this affect the public API?

src/node.h
nwellnhof: The C standard mandates a single implementation-defined type for `enum`s. Compilers typically use `int`.

nwellnhof:
> Query: does anything in this affect the public API?

Yes, the addition of `malloc` in `cmark_mem` in the last commit. But this only affects people who use custom allocators which seems like a rarely used feature.

jgm:
OK, can you put something very prominent in the commit message so I'll remember to highlight this API change in the next release?

src/node.h
jgm: oh, that's good to know.

nwellnhof:
I changed the pull request to use `realloc` instead of `malloc` so now there aren't any changes to the public API.

jgm:
Excellent, thanks.

jgm:
We're getting a new fuzzing error:
```
  | AddressSanitizer:DEADLYSIGNAL
-- | --
  | =================================================================
  | ==1==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000 (pc 0x081024b4 bp 0xff98d2e8 sp 0xff98ceb0 T0)
  | ==1==The signal is caused by a READ memory access.
  | ==1==Hint: address points to the zero page.
  | #0 0x81024b4 in __interceptor_strcmp /src/llvm-project/compiler-rt/lib/sanitizer_common/sanitizer_common_interceptors.inc:443:25
  | #1 0x82478b1 in is_autolink cmark/src/commonmark.c:149:10
  | #2 0x82478b1 in S_render_node cmark/src/commonmark.c:420:9
  | #3 0x824ff5e in cmark_render cmark/src/render.c:172:10
  | #4 0x8244a94 in cmark_render_commonmark cmark/src/commonmark.c:479:10
  | #5 0x819b608 in LLVMFuzzerTestOneInput cmark/test/cmark-fuzz.c:24:10
  | #6 0x809fe66 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned int) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556:15
  | #7 0x808c313 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned int) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:292:6
  | #8 0x8091a18 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned int)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:774:9
  | #9 0x80b6827 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19:10
  | #10 0xf7cd5636 in __libc_start_main
  | #11 0x80672e8 in _start
```

Since it just popped up, I suspect it has to do with these changes.
Can you investigate?

nwellnhof:
Should be fixed with https://github.com/commonmark/cmark/pull/329.

If it's OK, could you add me to the OSS-Fuzz auto_ccs?

jgm:
I'd be happy to add you, but I can't figure out how!

nwellnhof:
You'd have to submit a pull request to https://github.com/google/oss-fuzz, adding me to `auto_ccs` here: https://github.com/google/oss-fuzz/blob/master/projects/cmark/project.yaml. But I can submit the pull request myself and mention this thread.

jgm:
Sure, go ahead!

PR #328, opened 2020-01-24
build: substitute the path into the generate files

compnerd:
This resorts to the variable substitution to ensure the path embedded is
correct.  Without this, the path at the time of the configuration.  In
the case of the Swift project, this ended up searching in the *source*
directory rather than the *build* directory.  This will ensure that we
export the file to an absolute location and we use the same location in
the `cmarkConfig.cmake` file by means of CMake's `configure_file`
subsitution.


compnerd:
@nwellnhof - guess that answers the question: yes, it is better to substitute the path.
@jgm sorry about the churn for this

PR #346, opened 2020-06-24
Fix handling of empty strings when creating XML/HTML output.

steffen-kiess:
The API (e.g. `cmark_node_set_url()` or `cmark_node_set_on_exit()`) will store empty strings as `NULL` values. `cmark_render_xml()` however cannot handle `NULL` values in some cases, `cmark_render_html()` cannot handle `NULL` values as link targets. Fix this by checking for `NULL` before calling `strlen()`.


jgm:
good, thanks!

PR #356, opened 2020-08-24
define whitespace per spec

kivikakk:
The spec reads:

> A whitespace character is a space (`U+0020`), tab (`U+0009`), newline (`U+000A`), line tabulation (`U+000B`), form feed (`U+000C`), or carriage return (`U+000D`).

This PR updates `cmark_ctype.c` to respect this definition — currently only space/tab/newline/carriage return are, and line tabulation and form feed are excluded.


jgm:
great, thanks

Issue #354, opened 2020-08-22
Issue revealed by fuzzer


jgm:
See https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21218

oss-fuzz Issue 21218

Detected at commit
https://github.com/commonmark/cmark/commit/74e8f638ad1b5f259208e07621255a9e098cc4f3
but seems to persist after that commit is reverted

nwellnhof:
This also seems to be an issue with quadratic growth, this time related to reference links. The fuzzer also figured out that a `\x00` byte in URLs expands to an escaped Unicode replacement character `%EF%BF%BD` for an additional 9x expansion. Here's a simple demonstration:

    $ python3 -c 'print("[1] "*5000,"\n\n[1]: urn:","\x00"*20000,"\n", sep="")' |wc -c
    40013
    $ python3 -c 'print("[1] "*5000,"\n\n[1]: urn:","\x00"*20000,"\n", sep="")' |build/src/cmark |wc -c
    900105007

That's 900MB of output from a 40KB input document. If you go a little larger, the input buffer limit is triggered.

A quick fix is to limit the length of URLs. This at least avoids quadratic growth, but still allows a large linear expansion factor.

A better fix is to check the amount of generated replacements against the amount of data parsed, and abort if a certain expansion factor is exceeded. That's how XML parsers guard against billion laughs attacks.

(My commit only limited the internal buffer size, so this issue is caught earlier.)

jgm:
@nwellnhof Does PR #357 close this issue or is there something further?

nwellnhof:
It should close the issue, but let's wait for confirmation from OSS-Fuzz.

nwellnhof:
This is now verified as fixed: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21218#c95

jgm:
Great.

PR #357, opened 2020-08-24
Avoid quadratic output growth with reference links

nwellnhof:
Keep track of the number bytes added through expansion of reference
links and limit the total to the size of the input document. Always
allow a minimum of 100KB.

Unfortunately, cmark has no error handling, so all we can do is to
stop expanding reference links without returning an error. This should
never be an issue in practice though. The 100KB minimum alone should
cover all real-world cases.

See issue #354.


jgm:
sorry for the delay in merging!

PR #358, opened 2020-09-02
add support for groff mom filter

KatolaZ:
Hello, I added a simple filter for the groff mom package. The filter uses mom macros for most of the NODEs, and produces nicer output than man for letters or manuscript-like md sources.


KatolaZ:
checking why the test is failing and will resubmit

PR #359, opened 2020-09-02
add support for groff mom filter

KatolaZ:
Add support for groff mom macro output, which produces slightly better pdf files than the groff man filter. Modified from the original man.c


KatolaZ:
Sorry but I can't reproduce the test error (have no access to a macos machine). Hope you can provide more details on the specific issue, and why it is triggered? Thanks.

PR #360, opened 2020-09-23
Fix SR-13534: fix normalize_html test

Interfere:
Jira Link:
---
https://bugs.swift.org/browse/SR-13534

Description
---
The cause of the hang is unknown, but [this issue in CTest](https://gitlab.kitware.com/cmake/cmake/-/issues/20116) might be somehow related. The aim of this changes is to avoid using doctest.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -26,8 +26,7 @@ endif(WIN32)
 IF (PYTHONINTERP_FOUND)
 
   add_test(html_normalization
-    ${PYTHON_EXECUTABLE} "-m" "doctest"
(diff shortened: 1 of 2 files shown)


Interfere:
@jgm Could you please review. Thanks!

jgm:
Can you describe more fully what the issue is and what you've done to deal with it?  Save me the time of going through all the upstream issues and trying to puzzle it out.

Interfere:
I think I have opened the PR before checked everything. Apparently, the issue is not with the test itself. Will do more investigation, and if it is the case, open another PR with fixes. Thank you for your time.

PR #372, opened 2020-12-12
Remove unnecessary comment in output

danger89:
When calling the `cmark_render_commonmark` with a cmark_node*, which converts the AST to markdown text again, it should just give me the barebone markdown format, without _unnecessary_ comments.

This is the only comment I could find in the output.

Regards,

Melroy van den Berg

### DIFF
--- a/src/commonmark.c
+++ b/src/commonmark.c
@@ -218,7 +218,6 @@ static int S_render_node(cmark_renderer *renderer, cmark_node *node,
       // this ensures that a following indented code block or list will be
       // inteprereted correctly.
       CR();
-      LIT("<!-- end list -->");
       BLANKLINE();
     }
     break;


danger89:
Is this really causing regression issues??

danger89:
I think it can be solved by using:

```
- foo
- bar
+ baz
```

instead of:

```
- foo
- bar
<!-- end list -->
- baz
```

jgm:
The comment isn't u*nnecessary*, because without it you have one list instead of two.

It's true that there might be other ways of achieving the same effect, but this use of HTML comments is a standard way to separate lists in Markdown.  If we tried the approach of changing the list marker, the output would be less "portable," because many Markdown interpreters don't recognize a change of list marker as starting a new list.

danger89:
Ow I see. Well too bad, in that case I will just close this pr.

PR #374, opened 2020-12-31
Modernize CMake

Symbitic:
This pull request updates the cmake build system to use more modern features. The minimum required version is v3.12, which is found in all three GitHub runners.

Changes:
+ Replaced global commands like `add_compile_options` with `target_compile_options`. CMake strongly encourages encourages this practice.
+ Add a new option: `CMARK_EXE`. When turned off, the cmark executable will not be built. Useful for when building cmark as a subproject.
+ Export cmark_static as just cmark when shared libraries are disabled.
+ Update install directory for cmake files on Windows (Windows does not search lib/cmake/cmark).


Mikenux:
Hello!

In Evince (Document Viewer), there is work in progress to support markdown files (via md -> html -> pdf conversion; see https://gitlab.gnome.org/GNOME/evince/-/merge_requests/391).

I think the build option CMARK_EXE is what is needed to only build cmark as library for a flatpak build. So, it would be nice to have this. Thanks!

jgm:
@nwellnhof is there no build configuration currently accessible that allows building without creating the executable?

It has been a while; I don't recall all the reasons this wasn't merged, but I think they include:

- It's a very big patch, and hard for someone like me who isn't an expert on cmake to evaluate
- At the time we were considering other incompatible cmake patches
- We didn't at the time want to require cmake 3.12; in fact we've only recently increased the minimum cmake version required to 3.

I see the value in something like CMARK_EXE.  Maybe it's possible to extract that part of this patch without requiring more than cmake 3.7?

PR #378, opened 2021-03-09
Fix quadratic behavior when parsing inlines

### LINKED ISSUE #373: Quadratic behavior with smart quotes
Found by OSS-Fuzz: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28805

Reduced test case:

    python3 -c 'print("- "*7000+"s"+" -"*7000)' |build/src/cmark --smart >/dev/null

nwellnhof:
The inline parsing code would call cmark_node_append_child to append
nodes. This public function has a sanity check which is linear in the
depth of the tree. Repeated calls could show quadratic behavior in
degenerate trees. Use a special function to append nodes without this
check.

Fixes #373. Found by OSS-Fuzz.

### DIFF
--- a/src/node.c
+++ b/src/node.c
@@ -23,20 +23,21 @@ static CMARK_INLINE bool S_is_inline(cmark_node *node) {
 }
 
 static bool S_can_contain(cmark_node *node, cmark_node *child) {
-  cmark_node *cur;
-
-  if (node == NULL || child == NULL) {
+  if (node == NULL || child == NULL || node == child) {
     return false;
   }
 
-  // Verify that child is not an ancestor of node or equal to node.
-  cur = node;
-  do {
-    if (cur == child) {
-      return false;
+  // Verify that child is not an ancestor of node.
+  if (child->first_child != NULL) {
+    cmark_node *cur = node->parent;
+
+    while (cur != NULL) {
+      if (cur == child) {
+        return false;
+      }
+      cur = cur->parent;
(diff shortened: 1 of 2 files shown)


src/node.c
```diff
@@ -23,20 +23,21 @@ static CMARK_INLINE bool S_is_inline(cmark_node *node) {
...
 
 static bool S_can_contain(cmark_node *node, cmark_node *child) {
-  cmark_node *cur;
-
-  if (node == NULL || child == NULL) {
+  if (node == NULL || child == NULL || node == child) {
     return false;
   }
 
-  // Verify that child is not an ancestor of node or equal to node.
```
jgm, line 32: I can see why, if child has no children, it cannot be an ancestor of node.
But what is replacing the check to verify that child is not equal to node?

jgm:
Sorry it took me a while to look at this. It looks great.
I had one question (inline above).
Did you measure the effects of these changes on benchmarks?  I would imagine it speeds things up.

src/node.c
nwellnhof: The equality test is now a few lines above.

nwellnhof:
I didn't measure anything, but the changes should also speed up parsing real-world documents.

jgm:
Excellent.

jgm:
I ran the benchmarks and it does look like we get a small speedup (~2-3%).

Issue #384, opened 2021-07-01
Build fails on Debian Jessie


mlocati:
It seems that cmark 0.30.0 can't be compiled on Debian Jessie (compilation of cmark 0.29.0 succeeds):

From within a Docker image launched with `docker run --rm -it debian:jessie bash`:

```sh
$ apt-get update && apt-get install -qy cmake curl g++ python3
[...]
$ cd "$(mktemp -d)"
$ curl -sSL -o - https://github.com/commonmark/cmark/archive/0.30.0.tar.gz | tar xz
$ cd cmark-*
$ make -s -j$(nproc) cmake_build
-- The C compiler identification is GNU 4.9.2
-- The CXX compiler identification is GNU 4.9.2
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Performing Test HAVE_FLAG_ADDRESS_SANITIZER
-- Performing Test HAVE_FLAG_ADDRESS_SANITIZER - Failed
-- Performing Test HAVE_FLAG_SANITIZE_ADDRESS
-- Performing Test HAVE_FLAG_SANITIZE_ADDRESS - Success
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY - Success
-- Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY
-- Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY - Success
-- Performing Test COMPILER_HAS_DEPRECATED_ATTR
-- Performing Test COMPILER_HAS_DEPRECATED_ATTR - Success
-- Looking for stdbool.h
-- Looking for stdbool.h - found
-- Performing Test HAVE___BUILTIN_EXPECT
-- Performing Test HAVE___BUILTIN_EXPECT - Success
-- Performing Test HAVE___ATTRIBUTE__
-- Performing Test HAVE___ATTRIBUTE__ - Success
-- Found PythonInterp: /usr/bin/python3 (found suitable version "3.4.2", minimum required is "3")
-- Configuring done
[...]

PR #385, opened 2021-07-06
build: allow building with CMake[<3.3]

compnerd:
The `COMPILE_LANGUAGE` generator expression was introduced in CMake 3.3.
Debian Jessie has an older CMake version bundled, which prevents
building on that platform without manually building CMake.  This
uses a more error-prone mechanism of requiring the commonmark developer
to explicitly invoke `cmark_add_compile_options` on the target to ensure
that the flags are propagated for the targets without using the
generator expression.  Adding the options unconditionally prevents using
CMark in a build with a non-C programming language.

Fixes: #384

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,33 +28,34 @@ if(CMARK_STATIC AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
   SET(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS)
 endif()
 
-# Compiler flags
-if(MSVC)
-  # Force to always compile with W4
-  add_compile_options($<$<COMPILE_LANGUAGE:C>:/W4>)
-  add_compile_options($<$<COMPILE_LANGUAGE:C>:/wd4706>)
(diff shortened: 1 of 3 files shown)


jgm:
> requiring the commonmark developer to explicitly invoke cmark_add_compile_options on the target to ensure that the flags are propagated for the targets without using the generator expression

Not sure what you mean by this...do you mean that the standard build instructions won't work any more (something else needs to be done?)

compnerd:
> Not sure what you mean by this...do you mean that the standard build instructions won't work any more (something else needs to be done?)

Correct.  You will find that every instance of `add_executable` and `add_library` (except the aliases) are now followed by a call to the new function.  Because it is a custom function, this is something that isn't in the official CMake documentation and needs to be enforced by the project.  This is something that anyone adding new libraries or binaries to this project would need to be aware of.

jgm:
I don't think I made myself clear enough. What I'm asking about is whether the procedure for a user to build and install this library using cmake would be any different as a result of this change?

compnerd:
Oh, from the user's perspective everything remains the same as before.

jgm:
Thanks.

Issue #389, opened 2021-07-13
Quadratic behavior when parsing emphasis


nwellnhof:
Parsing a `*t ` sequence followed by a `_t*_ ` sequence exhibits quadratic behavior:

```
$ for n in 5000 10000 20000; do python3 -c "print('*t '*$n+'_t*_ '*$n)" |time -f "$n elems: %e secs" build/src/cmark >/dev/null; done
5000 elems: 0.19 secs
10000 elems: 0.88 secs
20000 elems: 5.45 secs
```

### Sample pattern

```
*t *t *t *t _t*_ _t*_ _t*_ _t*_
```

### Analysis

* After matching a `*` closer, the preceding `_` opener is removed.
* After parsing a `_` closer, no opener is found. The corresponding `openers_bottom` entry is set to preceding `*` opener.
* After matching the next `*` closer, the corresponding opener is removed. This is the entry in `openers_bottom`.
* The entry in `openers_bottom` becomes a dangling pointer.
* Subsequent searches for an opener for `_` closers will test against the dangling pointer (which is undefined behavior).
* Since the opener was removed, there will never be a match.
* The `openers_bottom` optimization stops working, resulting in quadratic behavior with abusive input patterns.

### Possible solutions

* Adjust `openers_bottom` when removing delimiters (somewhat expensive).
* Number the delimiters and store the delimiter position instead of a pointer in `openers_bottom`.

PR #390, opened 2021-07-13
Fix quadratic behavior when processing emphasis

### LINKED ISSUE #388: Quadratic behavior when parsing smart quotes
Simply parsing a sequence of single or double quote characters exhibits quadratic behavior when the "smart" option is enabled:

```
$ for n in 5000 10000 20000; do python3 -c "print(\"'\"*$n)" |time -f "$n chars: %e secs" build/src/cmark --smart >/dev/null; done
5000 chars: 0.15 secs
10000 chars: 0.70 secs
20000 chars: 4.33 secs
```

This seems to be caused by [this loop in `process_emphasis`](https://github.com/commonmark/cmark/blob/023d887d9fa65f8680bf7a471191dc161f210244/src/inlines.c#L678).

nwellnhof:
Fixes #388 and #389.

### DIFF
--- a/src/inlines.c
+++ b/src/inlines.c
@@ -33,6 +33,7 @@ typedef struct delimiter {
   struct delimiter *previous;
   struct delimiter *next;
   cmark_node *inl_text;
+  bufsize_t position;
   bufsize_t length;
   unsigned char delim_char;
   bool can_open;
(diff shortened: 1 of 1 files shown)


jgm:
Excellent fix!

PR #392, opened 2021-08-13
build: impertively define output name for static library

compnerd:
When building with an older CMake, the generator expression is not
evaluated properly and embedded into the final output name which is
incorrect.  This should repair the ability to build with an older CMake.

PR #393, opened 2021-08-21
Update CMakeLists.txt

compnerd:
Bump the minimum required CMake to 3.7.  CMake 3.1 is required for the `CMAKE_C_STANDARD` and `CMAKE_C_STANDARD_REQUIRED` options to take effect.  Relying on CMake to control the compiler this way allows us to avoid having to rebuild the logic for detecting the appropriate flags and ensuring that the compiler is able to support the necessary language version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.7)
 project(cmark VERSION 0.30.1)
 
 include("FindAsan.cmake")


compnerd:
@jgm this is a follow up from the discussion that occurred in #392.  By bumping the CMake version, we can ensure that the compiler is properly used by the build.

jgm:
Are there associated changes you'd make if the version is bumped?
Maybe they should be added here?
Otherwise I don't see why we should gratuitously bump the version; that just makes it not work for certain users without providing any advantage to anyone.

compnerd:
No, the desired change is already in the tree here:
https://github.com/commonmark/cmark/blob/master/CMakeLists.txt#L17-L19

Id say that we could bump to 3.1 if you prefer, but it seems unnecessarily low.  I would actually suggest that we should bump to 3.18 and then roll back to the more declarative based CMake with generator expressions.  However, this change conservatively bumps the version to ensure that the compiler flags for the language version are setup properly.

jgm:
got it, thanks

PR #425, opened 2021-09-24
Fix parsing of emphasis before links

### LINKED ISSUE #424: Parser incorrectly reading strong/emph in paragraphs that contain links.
The following input produces an incorrect syntax tree:
```C
const char* sampleInput = "*text* [link](google.com)";
cmark_node* cmark_root = cmark_parse_document(sampleInput, strlen(sampleInput), CMARK_OPT_DEFAULT);
const char* html = cmark_render_html(cmark_root, 0);
```

This produces a tree:
```html
<p>*text* <a href="google.com">link</a></p>
```

I would expect text to have emphasis tags.  It seems to only miss the last set of emphasis before the link.  try `"*text* *text2* [link](google.com)"`

Version: `"0.30.1"`
Platform: Windows, Visual Studio 2017 v15.9.33
config.h: 
 ```C
#ifndef CMARK_CONFIG_H
#define CMARK_CONFIG_H

#ifdef __cplusplus
extern "C" {
#endif

#define HAVE_STDBOOL_H

#ifdef HAVE_STDBOOL_H
  #include <stdbool.h>
#elif !defined(__cplusplus)
  typedef char bool;
#endif

/* #undef HAVE___BUILTIN_EXPECT */

/* #undef HAVE___ATTRIBUTE__ */

#ifdef HAVE___ATTRIBUTE__
  #define CMARK_ATTRIBUTE(list) __attribute__ (list)
#else
  #define CMARK_ATTRIBUTE(list)
#endif

#ifndef CMARK_INLINE
  #if defined(_MSC_VER) && !defined(__cplusplus)
    #define CMARK_INLINE __inline
  #else
    #define CMARK_INLINE inline
  #endif
#endif

/* snprintf and vsnprintf fallbacks for MSVC before 2015,
   due to Valentin Milea http://stackoverflow.com/questions/2915672/
*/

#if defined(_MSC_VER) && _MSC_VER < 1900

#include <stdio.h>
#include <stdarg.h>

#define snprintf c99_snprintf
#define vsnprintf c99_vsnprintf

CMARK_INLINE int c99_vsnprintf(char *outBuf, size_t size, const char *format, va_list ap)
{
    int count = -1;

    if (size != 0)
        count = _vsnprintf_s(outBuf, size, _TRUNCATE, format, ap);
    if (count == -1)
        count = _vscprintf(format, ap);

    return count;
}

CMARK_INLINE int c99_snprintf(char *outBuf, size_t size, const char *format, ...)
{
    int count;
    va_list ap;

    va_start(ap, format);
    count = c99_vsnprintf(outBuf, size, format, ap);
    va_end(ap);

    return count;
}

#endif

#ifdef __cplusplus
}
#endif

#endif
```

```

nwellnhof:
Fix a regression introduced with commit ed0a4bf. Also test the first
delimiter against stack_bottom in process_emphasis. This happened to
work with the old code and only resulted in an unnecessary scan.

Fixes #424.

### DIFF
--- a/src/inlines.c
+++ b/src/inlines.c
@@ -640,7 +640,8 @@ static cmark_node *handle_period(subject *subj, bool smart) {
 }
 
 static void process_emphasis(subject *subj, bufsize_t stack_bottom) {
-  delimiter *closer = subj->last_delim;
+  delimiter *candidate;
+  delimiter *closer = NULL;
   delimiter *opener;
   delimiter *old_closer;
   bool opener_found;
@@ -650,10 +651,10 @@ static void process_emphasis(subject *subj, bufsize_t stack_bottom) {
                                  stack_bottom, stack_bottom, stack_bottom};
 
   // move back to first relevant delim.
-  while (closer != NULL &&
-         closer->previous != NULL &&
-         closer->previous->position >= stack_bottom) {
-    closer = closer->previous;
+  candidate = subj->last_delim;
+  while (candidate != NULL && candidate->position >= stack_bottom) {
+    closer = candidate;
+    candidate = candidate->previous;
   }
 
   // now move forward, looking for closers, and handling each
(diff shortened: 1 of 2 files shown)


jgm:
Excellent, thanks.

PR #427, opened 2021-11-10
[Bugfix] Fix the backticks bug

Kyle-Ye:
See more info and context here https://github.com/apple/swift-cmark/pull/26

In brief:
When parse "`A" and "``A"
Before the PR: We'll get text sourcepos 1:2-1:2 and 1:3-1:3
After the PR: We'll get text sourcepos 1:1-1:2 and 1:1-1:3

### DIFF
--- a/api_test/main.c
+++ b/api_test/main.c
@@ -1021,6 +1021,26 @@ static void source_pos_inlines(test_batch_runner *runner) {
     free(xml);
     cmark_node_free(doc);
   }
+  {
+    static const char markdown[] =
+      "` It is one backtick\n"
+      "`` They are two backticks\n";
+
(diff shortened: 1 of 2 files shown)


jgm:
Looks good -- I'd only ask that we have `init_pos` instead of `initPos` for consistency with the style of the rest of the code.

Kyle-Ye:
> Looks good -- I'd only ask that we have `init_pos` instead of `initPos` for consistency with the style of the rest of the code.

https://github.com/commonmark/cmark/blob/a8da5a2f252b96eca60ae8bada1a9ba059a38401/src/inlines.c#L394-L397

Seems like initpos would better fit the position here.(startpos and endpos)

jgm:
Yes, initpos would be better.

Kyle-Ye:
> Yes, initpos would be better.

Done.

Kyle-Ye:
Could we run test on it again and merge it. cc @jgm

PR #294, opened 2019-04-23
Add built cmark.exe as an AppVeyor artifact

JerwuQu:
As requested in #291

Issue #291, opened 2019-03-30
Include build artifacts in AppVeyor builds


JerwuQu:
Since you don't seem to provide pre-built binaries for Windows already, having them be created as artifacts through AppVeyor (for which you already have existing build steps) would be nice.

jgm:
No objections. Want to send a PR?

JerwuQu:
Done in #294, closing.

PR #428, opened 2021-11-14
Add Windows artifact to CI

JerwuQu:
This is for #291 since #294 wasn't carried over in the switch over to GitHub Actions.

The binary could also be included in future releases on GitHub if wished.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,3 +100,8 @@ jobs:
         chcp 65001
         nmake.exe /nologo /f Makefile.nmake test
       shell: cmd
+    - name: Upload artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: cmark ${{ matrix.cmake_opts }}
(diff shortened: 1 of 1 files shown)


nwellnhof:
For the shared build, it doesn't make sense to publish cmark.exe without cmark.dll.

jgm:
Yes, I know.  I tried to add an `if:` clause to that step to exclude it, but it didn't work; I don't think I understand how these are meant to work.

I guess an alternative would be to include cmark.dll as well.  (Same clause for both builds, so we'd probably need to remove the thing that says "produce an error if the file is not found".)

nwellnhof:
Something like `if: ${{ matrix.build_type == 'static' }}`?

jgm:
That's what I tried but it didn't seem to work.

jgm:
I will try again.

jgm:
Worked this time -- I must have tried something slightly different last time.
Thanks.

PR #430, opened 2021-11-29
Fix *cmark_get_default_mem_allocator(void) prototype

cirnatdan:
Fixes `error: function declaration isn't a prototype [-Werror=strict-prototypes] `


jgm:
Thanks!

PR #459, opened 2023-01-23
Address implicit integer truncations

anticomputer:
This commit changes size integers to consistently be of type size_t to prevent implicit integer truncations during integer arithmetic.

What remain are explicit truncations to (bufsize_t) which is of type int32_t. This requires additional careful consideration.


anticomputer:
Closed. Targeted wrong branch for PR.

PR #460, opened 2023-01-28
chore: Add dependabot for GitHub Actions

nschonni:
Will keep things like `actions/checkout@v1` up-to-date, as V3 is the current version.
Filtered out the CI jobs so that the dependabot branches doen't build on push, but only on the PRs that it opens.
The `macos-latest` casing was because the schema validation in VSCode was flagging it, but it could be pulled out

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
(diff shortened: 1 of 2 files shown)


jgm:
Thanks!

PR #467, opened 2023-02-02
spec_tests.py: Add option to generate fuzz corpus

nwellnhof:
Add an option `--fuzz-corpus` that writes the test cases to separate files including the options header, so they can be used as seed corpus for fuzz testing.

### DIFF
--- a/test/spec_tests.py
+++ b/test/spec_tests.py
@@ -6,6 +6,7 @@ from difflib import unified_diff
 import argparse
 import re
 import json
+import os
 from cmark import CMark
 from normalize import normalize_html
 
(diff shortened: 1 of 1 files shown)


nwellnhof:
My plan is to use our own spec tests as seed corpus for OSS-Fuzz instead of an external test suite: https://github.com/google/oss-fuzz/blob/master/projects/cmark/build.sh

jgm:
Good idea.

nwellnhof:
Thanks! Here's the pull request for OSS-Fuzz: https://github.com/google/oss-fuzz/pull/9564

PR #471, opened 2023-04-01
Keep an incremental count of open block types

kevinbackhouse:
This is a rebase of the solution that we implemented in cmark-gfm to fix https://github.com/github/cmark-gfm/security/advisories/GHSA-66g8-4hjf-77xh.

The idea is to incrementally keep an up-to-date count of the number of "open" blocks, categorized by node type, so that the loop in `check_open_blocks` can bail out early when it can see from the counts that the rest of the list doesn't contain any nodes that need to be checked.

I've fuzzed this solution extensively on cmark-gfm, but not on cmark.

Reproduction steps for the bug that this PR fixes:

```bash
python3 -c 'n = 10000; print(" -" * n + "x" + "\n" * n)' | cmark
```

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -235,15 +235,18 @@ static void remove_trailing_blank_lines(cmark_strbuf *ln) {
 // Check to see if a node ends with a blank line, descending
 // if needed into lists and sublists.
 static bool S_ends_with_blank_line(cmark_node *node) {
-  if (S_last_line_checked(node)) {
-    return(S_last_line_blank(node));
-  } else if ((S_type(node) == CMARK_NODE_LIST ||
-              S_type(node) == CMARK_NODE_ITEM) && node->last_child) {
-    S_set_last_line_checked(node);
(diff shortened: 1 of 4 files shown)


kevinbackhouse:
Out of interest, I'm surprised that the checks ran automatically. I'm a first-time contributor to cmark, so normally would have to wait for approval to run the checks. Have you modified something in the settings to allow the checks to run for first-time contributors?

nwellnhof:
This fix is incomplete. Please don't merge yet.

@kevinbackhouse See my report at HackerOne.

jgm:
Yes, I've modified the settings so that no approval is needed except for new GH-ers.

kevinbackhouse:
I haven't seen the HackerOne report yet (I'm not on the team that has access), but I've independently figured out what the problem is. I think it means that this strategy of counting the number of open blocks won't work, and we have to go back to the original solution of putting an arbitrary limit on the nesting depth.

PR #472, opened 2023-04-01
Fix quadratic performance issue in list numbering

kevinbackhouse:
This is a cherry-pick of one of the commits that we added in cmark-gfm to fix https://github.com/github/cmark-gfm/security/advisories/GHSA-r8vr-c48j-fcc5. (Only one of the bugs affects cmark too.)

To reproduce the bug:

```bash
python3 -c 'n = 10000; print("1.\n" + " 2.\n"*n)' | time ./src/cmark -t commonmark
python3 -c 'n = 10000; print("1.\n" + " 2.\n"*n)' | time ./src/cmark -t man
```

The solution is to keep track of the current item number, rather than recomputing it on every iteration by counting the number of elements in the list.

### DIFF
--- a/src/commonmark.c
+++ b/src/commonmark.c
@@ -153,7 +153,6 @@ static bool is_autolink(cmark_node *node) {
 
 static int S_render_node(cmark_renderer *renderer, cmark_node *node,
                          cmark_event_type ev_type, int options) {
-  cmark_node *tmp;
   int list_number;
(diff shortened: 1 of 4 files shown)


kevinbackhouse:
It turns out that this solution is incorrect (see https://github.com/github/cmark-gfm/issues/321).
I'm going to implement a new solution and post a new PR when it's ready.

PR #473, opened 2023-04-02
Fix quadratic performance issue in list numbering

kevinbackhouse:
This is the same fix as https://github.com/github/cmark-gfm/pull/322. It fixes the bug described in https://github.com/github/cmark-gfm/security/advisories/GHSA-r8vr-c48j-fcc5.

To reproduce the bug:

```bash
python3 -c 'n = 10000; print("1.\n" + " 2.\n"*n)' | time ./src/cmark -t commonmark
python3 -c 'n = 10000; print("1.\n" + " 2.\n"*n)' | time ./src/cmark -t man
```

This PR replaces #472, which didn't work correctly (see https://github.com/github/cmark-gfm/issues/321).

### DIFF
--- a/src/cmark.h
+++ b/src/cmark.h
@@ -329,6 +329,17 @@ CMARK_EXPORT int cmark_node_get_list_tight(cmark_node *node);
  */
 CMARK_EXPORT int cmark_node_set_list_tight(cmark_node *node, int tight);
 
+/**
+ * Returns item index of 'node'. This is only used when rendering output
+ * formats such as commonmark, which need to output the index. It is not
+ * required for formats such as html or latex.
+ */
+CMARK_EXPORT int cmark_node_get_item_index(cmark_node *node);
+
+/** Sets item index of 'node'. Returns 1 on success, 0 on failure.
+ */
+CMARK_EXPORT int cmark_node_set_item_index(cmark_node *node, int idx);
+
 /** Returns the info string from a fenced code block.
  */
 CMARK_EXPORT const char *cmark_node_get_fence_info(cmark_node *node);
--- a/src/cmark.h
+++ b/src/cmark.h
@@ -329,17 +329,6 @@ CMARK_EXPORT int cmark_node_get_list_tight(cmark_node *node);
  */
 CMARK_EXPORT int cmark_node_set_list_tight(cmark_node *node, int tight);
 
-/**
- * Returns item index of 'node'. This is only used when rendering output
- * formats such as commonmark, which need to output the index. It is not
- * required for formats such as html or latex.
- */
-CMARK_EXPORT int cmark_node_get_item_index(cmark_node *node);
-
-/** Sets item index of 'node'. Returns 1 on success, 0 on failure.
- */
-CMARK_EXPORT int cmark_node_set_item_index(cmark_node *node, int idx);
-
 /** Returns the info string from a fenced code block.
  */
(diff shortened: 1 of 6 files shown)


jgm:
Thanks for submitting this!  I haven't had a chance to look at it in detail yet.

src/cmark.h
```diff
@@ -329,6 +329,17 @@ CMARK_EXPORT int cmark_node_get_list_tight(cmark_node *node);
...
+ * formats such as commonmark, which need to output the index. It is not
+ * required for formats such as html or latex.
+ */
+CMARK_EXPORT int cmark_node_get_item_index(cmark_node *node);
+
+/** Sets item index of 'node'. Returns 1 on success, 0 on failure.
+ */
+CMARK_EXPORT int cmark_node_set_item_index(cmark_node *node, int idx);
+
 /** Returns the info string from a fenced code block.
```
nwellnhof, line 343: Is there a reason to make these functions part of the public API?

kevinbackhouse: I was just copying the style of similar declarations in this file.

nwellnhof: The file `cmark.h` contains public symbols. Private symbols should be declared in one of the other header files. In this case, `node.h` seems like the best choice.

Personally, I'd simply set the struct member directly without a helper function. The whole thing is a bit of a hack because the rendering functions shouldn't modify nodes at all. But I don't see a better way to solve this problem, so I'd add a comment that the rendering code abuses the otherwise unused `list.start` member to store a temporary value.

nwellnhof:
This modifies the source document for no good reason. Why not store the current list index in a local variable?

kevinbackhouse:
> This modifies the source document for no good reason. Why not store the current list index in a local variable?

Sorry, it's been a really long time since I wrote this code and I don't remember all the details. The performance issue is caused by this code:

https://github.com/commonmark/cmark/blob/7d450d8c48d58603b55901d133cb0831f80bc89b/src/commonmark.c#L221-L225

It takes linear time to recompute the index, which is what causes the quadratic performance. I fixed it by caching the index on the node so that it doesn't need to be recomputed. As far as I can remember, I couldn't find a place to store it in a local variable, which is why I implemented it this way.

jgm:
We don't want to modify the source, but if there's a way to handle this with a local variable, great.

In general I haven't been too worried about performance-related issues in the commonmark or man renderer.  (HTML, yes.)  And the issue is one that will only arise if you have an unusually long list.

nwellnhof:
I just realized that for nested lists, you'd need a dynamic array to store list indexes. This would make things more complicated and adds a memory allocation to the rendering code which so far seems to be allocation-free. So I can see why the approach of storing indexes in the input document was chosen.

jgm:
The rendering code isn't allocation-free; `cmark_renderer` has two dynamic string buffers, `buffer` and `prefix`, as well as a dynamic linked structure `block_number_in_list_item` that gets used in the man writer.

kevinbackhouse:
Maybe a hashtable indexed by the node pointer? That would be functionally equivalent to storing data on the node itself, without actually modifying the node.

jgm:
Something very similar to the existing `block_number_in_list_item` should do the trick. That is also a kind of stack, in order to handle nested lists.  In addition to tracking the block number within the list item, we could track the list item item number. Perhaps the two features could even be integrated.

PR #483, opened 2023-09-23
Add copyright notices

bkmgit:
Include information of code derived from other sources

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -1,3 +1,25 @@
+/*
+ * buffer.h is derived from code (C) 2012 Github, Inc.
+ *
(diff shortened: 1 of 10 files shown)


bkmgit:
Solves https://github.com/commonmark/cmark/issues/505

@jgm Feedback on this would be appreciated. For further context please see https://bugzilla.redhat.com/show_bug.cgi?id=2227502

jgm:
Our copyright notice regarding utf8.c/h was added in commit 4e230226d1ffa3916fb7ae52a6b93fdf929ee850
@vmg represented this code as GitHub's; nothing was said about utf8proc.
If that is indeed the original source of the code, I'm happy to correct this mistake, but perhaps @vmg would like to comment.

PR #491, opened 2023-12-19
build: add an upgrade path for newer CMake

compnerd:
CMake 3.12 deprecated `FindPythonInterp`, and with CMake 3.27, were obsoleted with CMP0148.  Add a version check and switch to the new behaviour to allow building with newer releases.  When the minimum CMake version is increased, we can look at dropping the compatibility path.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -3,11 +3,21 @@
 # By default, we run the spec tests only if python3 is available.
 # To require the spec tests, compile with -DSPEC_TESTS=1
 
-if (SPEC_TESTS)
-  find_package(PythonInterp 3 REQUIRED)
-else(SPEC_TESTS)
-  find_package(PythonInterp 3)
-endif(SPEC_TESTS)
+if(SPEC_TESTS)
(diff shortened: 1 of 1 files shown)


jgm:
thanks

test/CMakeLists.txt
```diff
@@ -3,11 +3,21 @@
...
-  find_package(PythonInterp 3)
-endif(SPEC_TESTS)
+if(SPEC_TESTS)
+  set(PYTHON_REQUIRED REQUIRED)
+else()
+  set(PYTHON_REQUIRED)
+endif()
+
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.12)
+  find_package(Python3 ${PYTHON_REQUIRED} COMPONENTS Interperter)
```
nwellnhof, line 13: `Interperter` looks like a typo.

nwellnhof: Oh, I see it was fixed in a later commit.

compnerd: Yeah; I want to finish up some more cleanups and then see if we can add some additional coverage through GHA.

PR #495, opened 2023-12-22
build: clean up an obsoleted variable (NFC)

compnerd:
This variable was no longer in use and the recent cleanups had exposed it as being unnecessary.


jgm:
I'm not sure what is replacing this functionality?

compnerd:
It isn't being replaced at all - its just getting entirely dropped.  The variables are entirely unused, and neither roundtrip.cmd nor roundtrip.sh exist in the source tree.  I believe that it has been replaced by roundtrip_tests.py.

PR #503, opened 2023-12-31
replace usage of `CMARK_INLINE` with `inline`

compnerd:
Modern versions of MSVC (2015+) support most of the C99 standard, including the `inline` keyword.  Avoid the vendor extension `__inline` and use the ISO standard spelling.

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -41,7 +41,7 @@ static bool S_last_line_checked(const cmark_node *node) {
   return (node->flags & CMARK_NODE__LAST_LINE_CHECKED) != 0;
 }
 
-static CMARK_INLINE cmark_node_type S_type(const cmark_node *node) {
+static inline cmark_node_type S_type(const cmark_node *node) {
   return (cmark_node_type)node->type;
 }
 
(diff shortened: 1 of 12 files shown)


jgm:
@kainjow @nwellnhof Does this change make sense to you?
(cf. #76 for background)

nwellnhof:
Yes, if you want to drop support for older MSVC compilers.

You can also remove the `/TP` flag hack here: https://github.com/commonmark/cmark/blob/master/CMakeLists.txt#L50

compnerd:
@nwellnhof good point!  I'll do a follow up to clean that up as well.

jgm:
I don't know if there's any reason to support these old compilers. (But I don't know enough about the Windows world to answer that question.)

nwellnhof:
You can find a history of C99/C11 support in Microsoft compilers in my Stackoverflow answer: https://stackoverflow.com/a/28118893/1956010

It's a judgment call but sooner or later this cruft should be removed. MSVC 14.0 (Visual Studio 2015) should be able to compile libcmark out of the box.

jgm:
OK, let's do it sooner then!

PR #504, opened 2024-01-02
squelch C4232 warnings on MSVC builds

compnerd:
C4232 appertains to the identity of dllimported functions [1]. The address of dllimport'ed functions are not guaranteed to maintain identity as the address will be the address of the IAT thunk, which is module specific.  Two modules which bind to the same implementation may have different addresses. However, since the use of this is for `free`, it should be relatively safe as we do not expect to perform pointer identity comparisons.

[1] https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4232?view=msvc-170


compnerd:
With this change, I actually can build with VS2022 and no warnings!

PR #511, opened 2024-01-08
build: permit incremental linking on Windows

compnerd:
This was added with the very first commit that introduced CMake. There was no explanation for the incremental linking not being enabled. If this is a personal preference, the user can specify the flag via `-D CMAKE_SHARED_LINKER_FLAGS`. Although incremental linking does introduce padding in the binary for incremental links, that should not matter for release mode builds in theory as `/OPT:REF` and `/OPT:ICF` will trigger full links [1], which is the default in release mode.

[1] https://learn.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170


jgm:
I have no idea why that is there. (I only half understand CMake anyway; I think someone else contributed the orginal CMake build.)

Will leave this open for a bit in case anyone remembers.  @nwellnhof ?

nwellnhof:
Looks good to me.

PR #513, opened 2024-01-08
build: remove `CMARK_STATIC` and `CMARK_SHARED` options

compnerd:
Remove the `CMARK_STATIC` and `CMARK_SHARED` options as one of the two must be enabled always as the cmark executable depends on the library. Instead of having a custom flag to discern between the library type, use the native CMake option `BUILD_SHARED_LIBS` allowing the user to control which library to build [1]. This matches recommendations to only build a single copy of the library [2].

[1] https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html
[2] https://alexreinking.com/blog/building-a-dual-shared-and-static-library-with-cmake.html


nwellnhof:
Note that this is a breaking change which should be documented in the release notes.

I'd also suggest to add an option setting the default to "shared":

    option(BUILD_SHARED_LIBS "Build using shared libraries" ON)

compnerd:
Sure, happy to add a release note for this.  I don't really have an opinion on the default. I'm personally a fan of dynamic linking (that is what Linux distributions tend to prefer), though I felt like the project tended to prefer static linking. I can change the default to shared if that is preferred. CC: @jgm

jgm:
It's probably best if the default stays whatever it was before (static or shared, I don't even remember).

compnerd:
The tests seem to indicate `BUILD_SHARED=YES` generally, so I _think_ that the expectation was static linking. I don't honestly know, but @nwellnhof seems to prefer dynamic, so we could do that?

jgm:
I don't have strong feelings about this and happy to defer to the two of you.

nwellnhof:
The old default was to build both static and shared libraries. I agree that it's cleaner and less error-prone to only allow building one type of library. Regarding what should be built by default, I'd prefer shared libraries but that's just my opinion.

I think we should also add a bit of backward compatibility to not break the build for hundreds of users. See Github search results for example:

- https://github.com/search?q=cmark_shared%3Doff&type=code
- https://github.com/search?q=cmark_static%3Doff&type=code

PR #517, opened 2024-01-11
test: simplify test execution environment handling

compnerd:
Use generator expressions to compute the new path and avoid translations. This reduces complexity in the build and allows for a different build layout.


compnerd:
@jgm seems that only collaborators can re-trigger workflows. In this case, the linux failures seemed to be due to an apt failure and so should be re-triggered.

jgm:
Re-triggered them but still failing. Can try again later.

compnerd:
@jgm this should be good to merge I believe

PR #520, opened 2024-01-15
Fix two cases of quadratic behavior (GHSA-66g8-4hjf-77xh)

nwellnhof:
Fixes the two issues reported here: https://github.com/github/cmark-gfm/security/advisories/GHSA-66g8-4hjf-77xh

Also known as CVE-2023-24824.


nwellnhof:
The speedup is probably caused by the change to `S_insert_emph`.

jgm:
Yes - saw that!

PR #524, opened 2024-01-18
Allow parsing into an existing node

Ericson2314:
Variation on #522 for same purpose. A good deal smaller. First commit is implementation, second commit is test.

### DIFF
--- a/src/cmark.h
+++ b/src/cmark.h
@@ -467,9 +467,41 @@ CMARK_EXPORT
 cmark_parser *cmark_parser_new(int options);
 
 /** Creates a new parser object with the given memory allocator
+ *
+ * A generalization of `cmark_parser_new`:
+ * ```c
+ * cmark_parser_new(options)
+ * ```
+ * is the same as:
+ * ```c
+ * cmark_parser_new_with_mem(options, cmark_get_default_mem_allocator())
+ * ```
  */
 CMARK_EXPORT
 cmark_parser *cmark_parser_new_with_mem(int options, cmark_mem *mem);
+
+/** Creates a new parser object with the given node to use as the root
+ * node of the parsed AST.
+ *
+ * When parsing, children are always appended, not prepended; that means
+ * if `root` already has children, the newly-parsed children will appear
+ * after the given children.
+ *
+ * A generalization of `cmark_parser_new_with_mem`:
+ * ```c
+ * cmark_parser_new_with_mem(options, mem)
+ * ```
+ * is approximately the same as:
+ * ```c
+ * cmark_parser_new_with_mem_into_root(options, mem, cmark_node_new(CMARK_NODE_DOCUMENT))
+ * ```
+ *
+ * This is useful for creating a single document out of multiple parsed
+ * document fragments.
+ */
+CMARK_EXPORT
+cmark_parser *cmark_parser_new_with_mem_into_root(
+    int options, cmark_mem *mem, cmark_node *root);
 
 /** Frees memory allocated for a parser object.
  */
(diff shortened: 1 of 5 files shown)


Ericson2314:
@jgm @nwellnhof does this look better?

src/cmark.h
```diff
@@ -466,11 +466,24 @@ CMARK_EXPORT void cmark_consolidate_text_nodes(cmark_node *root);
 CMARK_EXPORT
 cmark_parser *cmark_parser_new(int options);
```
jgm, line 468: Can we have a comment here in the same style as with the other exported functions?

src/cmark.h
```diff
@@ -466,11 +466,24 @@ CMARK_EXPORT void cmark_consolidate_text_nodes(cmark_node *root);
...
  */
 CMARK_EXPORT
 cmark_parser *cmark_parser_new_with_mem(int options, cmark_mem *mem);
 
+/** Creates a new parser object with the given node to use as the root
+ * node of the parsed AST.
+ *
+ * This is useful for creating a single document out of multiple parsed
+ * document fragments.
+ */
```
jgm, line 482: Reading this, I wondered:  what happens if the specified root node already has children?  What happens then?  Do the new nodes get added to the end of the children?  Or do they replace existing children?  Maybe the comment should be explicit about this.

src/cmark.h
Ericson2314: oh oops that was a stray function I never ended up making.

Did add more docs though.

src/cmark.h
Ericson2314: They appear after, see `add_child` in `blocks.c`.

I do wish it and `cmark_node_append_child` shared code though to make that clearer, however.

Ericson2314:
I noticed that `cmark.3` was stale. I fixed it a a new preparatory commit, but i think it would be nice to either not commit the generated file, or have a CI action that ensures it is up to date.

jgm:
cmark.3 is updated as part of `make all`.
Since I use the Makefile for releases, it will always be updated before a release.

nwellnhof:
I would prefer an API like

    int cmark_parser_set_target(cmark_parser *parser, cmark_node *target);
    cmark_node *cmark_parser_get_target(cmark_parser *parser);

mainly because it makes it easier to support this feature in language bindings. For memory management, the reference from the parser object to the node object must be tracked and the `get_target` function provides easy access to this reference.

Besides, it would get more and more clumsy if we wanted to support additional settings in the parser constructor.

I'm aware that it's cleaner to avoid mutation of parser state and to set all options when constructing the parser object. Functions like `cmark_parser_set_target` should return an error if they're called after `cmark_parser_feed`.

Ericson2314:
What if we had a `cmark_parser_args` struct where everything was nullable // optional? Flexible (don't deal with args you don't care about) and also no new stateful-ness for the parser itself?

nwellnhof:
Exposing structs in public APIs should be avoided because modifying the struct in future versions would break the ABI. There are some tricks like adding a version field to the struct but that's rather ugly.

The cleanest solution is to have a separate object for parser settings, leading to something like

    cmark_parser_settings *settings = cmark_parser_settings_new();
    cmark_parser_settings_set_options(settings, CMARK_OPT_SMART);
    cmark_parser_settings_set_target(settings, target);
    cmark_parser *parser = cmark_parser_new_with_settings(settings);
    // ...
    cmark_parser_free(parser);
    cmark_parser_settings_free(settings);

For a real-world example look at `pthread_mutex_init ` and `pthread_mutexattr_t` in pthreads. But making such a change to the libcmark API seems a bit intrusive to me.

Ericson2314:
Ooo I like that. Making another opaque type to keep some state out of the parser feels good to me.

PR #529, opened 2024-01-29
Super-WIP take terminal render for lowdown

### LINKED ISSUE #362: Add support for ANSI output format
Currently, there is `man` output format (`-t, --to`). It can be viewed using something like

```console
$ cmark -t man < README.md | man -l -
```

Unfortunately, produced output is absolutely unreadable, *unless* it follows closely man pages conventions, eg, declares all required sections (**`NAME`**, **`SYNOPSIS`**, etc). This is very unfortunate situation when most of README's you're reading is from public git repositories.

However, displaying nicely formatted, colored Markdown in terminal is possible, using ANSI "formatting" sequences. Here is example of such representation:

image

(Colors are a bit weird, but this is just illustration)

So, it would be nice to have ANSI as supported output format. What do you think?

Ericson2314:
This doesn't work at all yet. It's just a sedded copy paste from https://github.com/kristapsdz/lowdown to close #362.

I'm just testing the waters here, per @jgm saying in https://github.com/commonmark/cmark/issues/362#issuecomment-1900820490 "I'd be open to it, but ...".

I will admit ~2000 lines more C for a more opinionated/open-ended feature (there is no way for a terminal rendering to be canonically correct, I would think) feels gross to me. (Maybe there are ways to deduplicate this with the rest of `cmark`, but I would not be confident in that. The rest of `lowdown` is vaguely similar and this wasn't deduplicated with that either.)

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -1,0 +1,1690 @@
+/*
+ * Copyright (c) Kristaps Dzonsons <kristaps@bsd.lv>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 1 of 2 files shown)


jgm:
> 2000 lines more C for a more opinionated/open-ended feature (there is no way for a terminal rendering to be canonically correct, I would think) feels gross to me. 

I'm inclined to agree with this.  It's good to keep this library lean.
Perhaps, for your purposes, you could develop this as a separate library, and then link it with cmark in your finished product.

Ericson2314:
Do you think there will be enough accessors functions to get out all the node-type-specific node info from the union?

jgm:
> Do you think there will be enough accessors functions to get out all the node-type-specific node info from the union?

You tell me.  I believe the API exposes full information about the AST, but if something is missing, let us know.

Ericson2314:
Great; knowing that it's at least *supposed* to expose enough stuff to support a downstream renderer --- and thus any gaps can be filled --- is good :).

PR #531, opened 2024-03-04
build: Tell cmake to set 'rpath'

mfwitten:
Before this commit, the `cmark` executable didn&rsquo;t inform the dynamic linker where to look for the `libcmark` shared object; this becomes an irritation when `libcmark` is installed in an unorthodox location:

<pre>
$ <a href="https://man7.org/linux/man-pages/man1/ldd.1.html">ldd</a> <b>/path/to/installed/files</b>/bin/cmark
        linux-gate.so.1 (0xb7ed7000)
        libcmark.so.0.31.0 => <b>not found</b>
        libc.so.6 => /usr/lib/libc.so.6 (0xb7cf3000)
        /lib/ld-linux.so.2 => /usr/lib/ld-linux.so.2 (0xb7ed8000)
</pre>

Because of this commit, the `cmark` executable&rsquo;s `rpath`<sup>\[[0](https://sourceware.org/binutils/docs-2.42/ld/Options.html#index-runtime-library-search-path)\]\[[1](https://cmake.org/cmake/help/v3.29/variable/CMAKE_INSTALL_RPATH.html)\]</sup> setting (or equivalent) is set upon installation, thereby providing the required search directory:

<pre>
$ <a href="https://man7.org/linux/man-pages/man1/ldd.1.html">ldd</a> <b>/path/to/installed/files</b>/bin/cmark
        linux-gate.so.1 (0xb7ef2000)
        libcmark.so.0.31.0 => <b>/path/to/installed/files/lib/libcmark.so.0.31.0 (0xb7e95000)</b>
        libc.so.6 => /usr/lib/libc.so.6 (0xb7cbc000)
        /lib/ld-linux.so.2 => /usr/lib/ld-linux.so.2 (0xb7ef3000)
</pre>

In particular, `rpath` is set to the following value:

<pre>"${<a href="https://cmake.org/cmake/help/v3.29/variable/CMAKE_INSTALL_PREFIX.html">CMAKE_INSTALL_PREFIX</a>}/${<a href="https://cmake.org/cmake/help/v3.29/module/GNUInstallDirs.html">CMAKE_INSTALL_LIBDIR</a>}"</pre>

Of course, this will only help on a system that supports an `rpath` feature known to `cmake`. Windows has no such feature anyway, apparently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,21 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
+if (DEFINED CMAKE_INSTALL_RPATH)
+  set(Base_rpath "${CMAKE_INSTALL_RPATH}")
+else()
+  if(BUILD_SHARED_LIBS)
+    set(p "${CMAKE_INSTALL_FULL_LIBDIR}")
+    list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${p}" i)
+    if("${i}" STREQUAL "-1")
+      set(Base_rpath "${p}")
+    endif()
+  endif()
+endif()
+
+# Append non-standard external dependency directories, if any.
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
 # Check integrity of node structure when compiled as debug
 add_compile_options($<$<CONFIG:Debug>:-DCMARK_DEBUG_NODES>)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -28,7 +28,6 @@ add_library(cmark
   xml.c)
 cmark_add_compile_options(cmark)
 set_target_properties(cmark PROPERTIES
-  MACOSX_RPATH TRUE
   OUTPUT_NAME "cmark"
   # Avoid name clash between PROGRAM and LIBRARY pdb files.
   PDB_NAME libcmark
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -55,7 +55,8 @@ add_executable(cmark_exe
   main.c)
 cmark_add_compile_options(cmark_exe)
 set_target_properties(cmark_exe PROPERTIES
-  OUTPUT_NAME "cmark")
+  OUTPUT_NAME "cmark"
+  INSTALL_RPATH "${Base_rpath}")
 target_link_libraries(cmark_exe PRIVATE
   cmark)
 


nwellnhof:
The rpath should be set to `${CMAKE_INSTALL_FULL_LIBDIR}` to make absolute libdirs work.

Then we should consider not adding an rpath for system library directories, see `FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES` in the [CMake Wiki](https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling). Alternatively, we could make this feature optional.

cmake/modules/SetRPATH.cmake
```diff
@@ -0,0 +1,15 @@
+set(CMAKE_SKIP_BUILD_RPATH FALSE)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
```
nwellnhof, line 2: Both of these variables should already be set to FALSE by default. I think they are only mentioned in the CMake Wiki for clarity.

cmake/modules/SetRPATH.cmake
```diff
@@ -0,0 +1,15 @@
...
+
+# Append non-standard external dependency directories
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```
nwellnhof, line 5: libcmark has no external dependencies, so this is unnecessary.

cmake/modules/SetRPATH.cmake
```diff
@@ -0,0 +1,15 @@
...
+
+# CMake 3.25 has a better scoping solution: 'block()'
+function(SetRPATH)
+  set(p "${CMAKE_INSTALL_FULL_LIBDIR}")
+  list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${p}" i)
+  if("${i}" STREQUAL "-1")
+    set(CMAKE_INSTALL_RPATH "${p}" PARENT_SCOPE)
+  endif()
+endfunction()
+SetRPATH()
```
nwellnhof, line 15: I'm not sure if we need a separate function.

cmake/modules/SetRPATH.cmake
mfwitten: Well, *exactly*. They are mentioned for clarity.

* Defaults are great, until they're not; convention is a terrible foundation.

* As it turns out, `cmake `has a lot of knobs related to `rpath`; one might as well set the knobs explicitly, and in one place, so as to help anybody in the future who has some *unforeseen* need.

Nevertheless:

* The reason it made sense to me to remove the explicit setting of the *target* property `MACOSX_RPATH` is that such an option never made much sense, anyway.

* If you still don&rsquo;t agree with setting some of these variables explicitly, then perhaps they can be simply commented out. Would that be acceptable?

cmake/modules/SetRPATH.cmake
mfwitten: Of course `libcmark` has external dependencies.

<pre>
$ <a href="https://man7.org/linux/man-pages/man1/ldd.1.html">ldd</a> <b>/path/to/installed/files</b>/bin/cmark
        linux-gate.so.1 (0xb7ef2000)
        libcmark.so.0.31.0 => /path/to/installed/files/lib/libcmark.so.0.31.0 (0xb7e95000)
        libc.so.6 => <b>/usr/lib/libc.so.6</b> (0xb7cbc000)
        /lib/ld-linux.so.2 => <b>/usr/lib/ld-linux.so.2</b> (0xb7ef3000)
</pre>

For example:
<pre>
cd "$repo"
d=/tmp/custom
p='
    {print}
    /^include\(GNUInstallDirs\)$/ {
      print("link_directories(\"'"$d"'\")")
    }
'
f=CMakeLists.txt; t=$f.tmp; <"$f" awk >"$t" "$p" && mv -f "$t" "$f"
git diff | cat
echo '================='
i='/path/to/installed/files'
make INSTALL_PREFIX="$i" install | grep 'Set runtime path'
</pre>

Here are the results of running those commands:

<pre>
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 054f8d4..22d9506 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ include(CMakePackageConfigHelpers)
 include(CTest)
 include(GenerateExportHeader)
 include(GNUInstallDirs)
<b>+link_directories("/tmp/custom")</b>
 
 if(NOT MSVC OR CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
   set(CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_NO_WARNINGS ON)
=================
-- Set runtime path of "/path/to/installed/files/bin/cmark" to "/path/to/installed/files/lib:<b>/tmp/custom</b>"
-- Set runtime path of "/path/to/installed/files/lib/libcmark.so.0.31.0" to "/path/to/installed/files/lib:<b>/tmp/custom</b>"
</pre>

Then:

<ul>
<li>
<b>Default to System Library</b>
<pre>
$ ldd "$i"/bin/cmark
        linux-gate.so.1 (0xb7ef2000)
        libcmark.so.0.31.0 => /path/to/installed/files/lib/libcmark.so.0.31.0 (0xb7e95000)
        libc.so.6 => <b>/usr/lib</b>/libc.so.6 (0xb7cbc000)
        /lib/ld-linux.so.2 => /usr/lib/ld-linux.so.2 (0xb7ef3000)
</pre>
</li>
<li>
<b>Custom Library</b>
<pre>
$ mkdir -p "$d"
$ cp -a -t "$d" /usr/lib/libc.so.6 /usr/lib/libc-2.29.so
$ ldd "$i"/bin/cmark
        linux-gate.so.1 (0xb7ef2000)
        libcmark.so.0.31.0 => /path/to/installed/files/lib/libcmark.so.0.31.0 (0xb7e95000)
        libc.so.6 => <b>/tmp/custom</b>/libc.so.6 (0xb7cbc000)
        /lib/ld-linux.so.2 => /usr/lib/ld-linux.so.2 (0xb7ef3000)
</pre>
</li>
</ul>

In other words:

* By simply adding `link_directories("/tmp/custom")` to the `CMakeLists.txt`, the directory `/tmp/custom` got added to the `rpath`.

* The only reason this works is because `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)`.

It's not unreasonable to think that at some point, in the future, an external dependency might be possible; maybe Nvidia will release a driver for offloading markdown parsing to the GPU. :-P

cmake/modules/SetRPATH.cmake
mfwitten: I would have much preferred to write:

<pre>
block()
  set(p "${CMAKE_INSTALL_FULL_LIBDIR}")
  list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${p}" i)
  if("${i}" STREQUAL "-1")
    set(CMAKE_INSTALL_RPATH "${p}" PARENT_SCOPE)
  endif()
endblock()
</pre>

However, the CMake language only got [anonymous] block scoping in version `3.25`, while this project currently sets `3.7` as the minimum version; I cannot use the `block()` construct.

This is why I added the comment:

    # CMake 3.25 has a better scoping solution: 'block()'

My hope was that it would be a hint to some future janitor to use the `block()` construct when available.

cmake/modules/SetRPATH.cmake
nwellnhof: I think it's misleading to set variables to their default values.

cmake/modules/SetRPATH.cmake
nwellnhof: Well, obviously the C library is a dependency but it doesn't seem useful to support this use case. Maybe we should also consider to only set the INSTALL_RPATH target property for the cmark executable.

cmake/modules/SetRPATH.cmake
nwellnhof: There are only five relevant Iines in that block and I think it's clearer to add them directly in CMakeLists.txt without adding an additional file.

cmake/modules/SetRPATH.cmake
mfwitten: Of course, it's not misleading *at all*. Quite the opposite.

However, I'll choose to interpret your reply as "Look, just remove them, OK?"

cmake/modules/SetRPATH.cmake
mfwitten: * Shared libraries have dependencies, too.

        $ ldd "$i"/lib/libcmark.so
                linux-gate.so.1 (0xb7fa3000)
                libc.so.6 => /tmp/custom/libc.so.6 (0xb7da3000)
                /usr/lib/ld-linux.so.2 (0xb7fa4000)

* **However**, it's straightforward enough to set only `cmark`&rsquo;s `rpath` to an unorthodox installation path, while still allowing `cmake` to set `libcmark`&rsquo;s `rpath` if it deems necessary (as described above); then, in most cases (pretty much *always*), `libcmark`&rsquo;s `rpath` will be the empty string, even when installed in an unorthodox location.

* The goal isn&rsquo;t to support *that* use case.

    (However, it would indeed *now* be supported; I mean, maybe someone wants to build against [*musl*](https://www.musl-libc.org/) or something, and has put everything in a separate, non-standard hierarchy.)

    Rather, the goal is to support a *logical policy* in general, namely:

    > If `cmake` is *ever* aware of an external dependency (or search directory) that is not covered by the standard locations, then `cmake` should add the relevant path to the `rpath`.

    That sounds like a sensible policy; it sounds like a sensible *default*.

cmake/modules/SetRPATH.cmake
mfwitten: Actually, it would be much clearer to move *even more stuff* into their own files, but I guess we have different tastes. I will move it all into `CMakeLists.txt`, as that is what you prefer.

mfwitten:
I think I have hit all the buttons this time. &#x1f91e;

* There is no new file.
* There is no function scope.
* There is no redundant default.
* No standard path is repeated in `rpath`.
* `libcmark` will essentially never have an `rpath`.
* Yet, `cmake` may still add paths to each `rpath` as necessary.

mfwitten:
I pushed a slightly modified version; as described in the commit log:

> If `CMAKE_INSTALL_RPATH` is set, then it is used; this allows the user to set a single, overriding value the expected way.

Essentially:

<pre>
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9170bf3..21c1763 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,9 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
<b>+if (DEFINED CMAKE_INSTALL_RPATH)</b>
<b>+  set(Base_rpath "${CMAKE_INSTALL_RPATH}")</b>
<b>+else()</b>
   if(BUILD_SHARED_LIBS)
     set(p "${CMAKE_INSTALL_FULL_LIBDIR}")
     list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${p}" i)
@@ -76,6 +79,7 @@ if(BUILD_SHARED_LIBS)
       set(Base_rpath "${p}")
     endif()
   endif()
<b>+endif()</b>
 
 # Append non-standard external dependency directories, if any.
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
</pre>

jgm:
Sorry, I forgot about this.  @nwellnhof if it still looks good to you I will merge this.

nwellnhof:
Looks good.

PR #534, opened 2024-03-18
ci: Test with ASan and UBSan

nwellnhof:
This adds a CI job that tests with clang, ASan and UBSan. Only the static build is supported for now, meaning that we miss a few tests.

I happened to run into a very recent and probably short-lived issue with Github runners: https://github.com/actions/runner-images/issues/9491. The work-around can probably be removed soon.

We could also consider removing the valgrind test now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,27 @@ jobs:
          ctest --test-dir build --output-on-failure
          make leakcheck
 
+  sanitizers:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    env:
+       CC: 'clang'
+       CXX: 'clang++'
+       CFLAGS: '-fsanitize=address,undefined -fno-sanitize-recover=all'
+       CXXFLAGS: '-fsanitize=address,undefined -fno-sanitize-recover=all'
+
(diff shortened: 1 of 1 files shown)


nwellnhof:
Maybe we should discuss this before merging.

Here's a more invasive change that enables sanitizers on all OSs and compilers: https://github.com/nwellnhof/cmark/blob/asan2/.github/workflows/ci.yml

jgm:
So, is the question here whether we want the minimal change or the one that tests with sanitizers on all OSs/compilers?  What is the drawback to the latter option?

nwellnhof:
The new proposal now updated with a full os/compiler/build_type/sanitizer matrix: https://github.com/nwellnhof/cmark/blob/asan2/.github/workflows/ci.yml

Test results: https://github.com/nwellnhof/cmark/actions/runs/8321495139

nwellnhof:
I'll post a new PR.

jgm:
Are there any drawbacks to using this full matrix?  I wasn't sure what you had in mind when you suggested discussion.

nwellnhof:
> Are there any drawbacks to using this full matrix?

Not really. The full matrix runs a few more jobs than before.

PR #535, opened 2024-03-18
ci: Test with ASan and rework CI matrix

nwellnhof:
Add CI tests with -fsanitize=address,undefined. This is a bit complicated since our tests run through Python FFI and we have to

- link with -shared-libasan
- preload the ASan runtime
- disable LeakSanitizer because of false positives from Python

Use a single matrix on POSIX with four human readable labels:

    posix (os, compiler, build_type, sanitizers)

Disable gcc on macOS which is just an alias for clang.

Remove the valgrind leakcheck test. This should be covered now by testing with LeakSanitizer on static builds, including tests like spectest_executable. The full test suite is only available to shared library builds which have to run without leak checks for now.

Also rework the Windows matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,67 +27,92 @@ jobs:
         CC: clang
         CXX: clang++
 
-  linux:
+  posix:
 
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        cmake_opts:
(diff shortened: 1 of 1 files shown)


nwellnhof:
Github action page: https://github.com/commonmark/cmark/actions/runs/8321937012

jgm:
Great!

PR #536, opened 2024-03-24
Misc cleanup

nwellnhof:


src/houdini.h
```diff
@@ -25,8 +25,6 @@ extern "C" {
 
 extern bufsize_t houdini_unescape_ent(cmark_strbuf *ob, const uint8_t *src,
                                       bufsize_t size);
-extern int houdini_escape_html(cmark_strbuf *ob, const uint8_t *src,
-                               bufsize_t size);
```
jgm, line 29: In addition to removing this, would it make sense to rename `houdini_escape_html0` to `houdini_escape_html`?

nwellnhof: Yes, this would make sense.

jgm: I've done this.

PR #537, opened 2024-03-24
Reduce binary size of the library

nwellnhof:
These two commits rework Unicode case folding and HTML entity lookup to reduce the size of the shared library from 345 KB to 202 KB on Ubuntu 22.04.


jgm:
Excellent!

PR #538, opened 2024-03-24
fuzz: Test more parser entry points

nwellnhof:
Use the upper bits of 'options' to select a parser mode from

- cmark_parse_document
- cmark_parse_file
- cmark_parser_new/feed/finish
- cmark_markdown_to_html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,9 +89,10 @@ afl:
 	    $(CMARK) $(CMARK_OPTS)
 
 libFuzzer:
-	CC=clang CXX=clang++ cmake -S . -B $(BUILDDIR) \
-	    -DCMAKE_BUILD_TYPE=Asan \
-	    -DCMARK_LIB_FUZZER=ON
+	CC=clang CXX=clang++ \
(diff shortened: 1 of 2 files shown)


jgm:
CIFuzz is failing after this.  Do you want me to merge it anyway?

nwellnhof:
It turned out that the libFuzzer Makefile target didn't enable ASan. I force pushed some fixes.

`posix (macos, clang, shared)` now failed because of a timeout. This seems unrelated.

nwellnhof:
I force-pushed again and all tests are green.

nwellnhof:
OSS-Fuzz coverage is now at 96%: https://storage.googleapis.com/oss-fuzz-coverage/cmark/reports/20240402/linux/src/cmark/src/report.html

My plan is to add a fuzzer for the uncovered tree API functions similar to work I recently completed on libxml2: https://gitlab.gnome.org/GNOME/libxml2/-/blob/master/fuzz/api.c?ref_type=heads. But this might take a while.

PR #543, opened 2024-04-11
Adding PHP library to the list

CViniciusSDias:


jgm:
Do I understand correctly from this that there's a commonmark parser built into PHP itself?

CViniciusSDias:
It's an external extension. I just found out about that one today. 😅 

Extensions in PHP are C code compiled and embedded into the VM, so they work like they are part of the language.

PR #546, opened 2024-04-12
Fix wrappers leaking memory

### LINKED ISSUE #544: Example wrappers leak memory
The Python, Ruby and PHP wrappers leak memory. The C string returned from `cmark_markdown_to_html` must be freed (by calling `free`) after converting to a host language string. I recently fixed this for `cmark.py` in dfad5a9.

nwellnhof:
Fixes #544


jgm:
Excellent, thanks.

PR #552, opened 2024-07-19
Makefile: instead of python, use python3

amirouche:


amirouche:
The build is broken on several distribution: https://github.com/letloop/letloop/actions/runs/10013488401

jgm:
My question is whether an executable named `python3` is always available when python v3 is installed.  It seems to be on macOS and Debian -- what about others?

tats-u:
@jgm 
Alma: available after installing `python3` package
Windows: built-in Microsoft Store redirector supports `python3` shim

```pwsh
PS> gcm python3 -all | % { $_.Source.Replace("$HOME", "~") }
~\.rye\shims\python3.exe
~\AppData\Local\Microsoft\WindowsApps\python3.exe
```

Ubuntu: Debian-based; 100% exists

PR #557, opened 2024-08-18
[cjk-branch] Tweak CJK handling

tats-u:
- Add Yijing Hexagram Symbols
- See one more previous character if the previous character is VS15 (U+FE0E; suppresses emoji rendering)
- Correct wrong Unicode block description


tats-u:
The test case for Yijing passes even with the current implementation.
This is why it was removed by the latest force push.

jgm:
Many thanks for this.

Issue #558, opened 2024-08-19
Consider allowing Nodes with a "fragment" type.


ioquatix:
I'm interested in manipulating markdown ASTs.

It's sometimes problematic to extract a fragment, e.g.

```ruby
header = Markly.parse("### Hello `World`").first_child
fragment = header.extract_children # Move all children to a node of type `custom`
fragment.to_markdown # => "Hello `World`\n\n"
```

I'm having trouble with the output formatting.

Ideally, the fragment would just be the raw children, e.g.

```ruby
fragment.to_markdown # => "Hello `World`"
```

This would make it a little easier to mash up the AST with different types of outputs.

I suppose the issue is the formatting of `custom` type nodes, but I'm not sure if there is a problem to change this.

In other words, either (1) we could change the output to be "inline" or add an inline "fragment" node type which does minimal formatting. Thoughts?

ioquatix:
Okay, I just realised there is an `custom_inline` type... I'm trying it now...

ioquatix:
So, even with `custom_inline`, I'm getting newline characters in the output:

```
fragment.to_markdown
=> "Hello `World`\n"
```

Not sure why.

ioquatix:
Okay, I found the problematic code:

```c
  // ensure final newline
  if (renderer.buffer->size == 0 || renderer.buffer->ptr[renderer.buffer->size - 1] != '\n') {
    cmark_strbuf_putc(renderer.buffer, '\n');
  }
```

I believe we should only invoke this if the top level node is a block.

ioquatix:
So there are probably two issues here:

- Why does using `custom_block` generate two trailing newlines - is that expected?
- Can we change `custom_inline` to not generate trailing newlines?
- If we can't change the default, can we introduce an option to control this, or perhaps another node type (`fragment`) for handling this case?

jgm:
The renderer is designed to ensure that the output produced by cmark ends with a newline.  Unix text files are supposed to end with a newline.

The `custom_block` renderer adds a newline after the custom block because new blocks in commonmark must always start on a new line.

Isn't it easy enough to strip off the newline in your application?

ioquatix:
As you said yourself, in some cases the newline is semantic, so I'm not sure stripping it off arbitrarily is a good idea.

The goal of a fragment is to be able to inject it correctly into existing markdown, e.g. moving a title text into a list item:

```ruby
node = Markly.parse("# My Title")
# Extract a fragment:
children = node.extract_children

[...]

PR #559, opened 2024-08-19
Don't append a newline character when rendering inline nodes.

ioquatix:
See <https://github.com/commonmark/cmark/issues/558> for context.

### DIFF
--- a/src/render.c
+++ b/src/render.c
@@ -178,9 +178,11 @@ char *cmark_render(cmark_node *root, int options, int width,
     }
   }
 
-  // ensure final newline
-  if (renderer.buffer->size == 0 || renderer.buffer->ptr[renderer.buffer->size - 1] != '\n') {
-    cmark_strbuf_putc(renderer.buffer, '\n');
+  // If the root node is a block type (i.e. not inline), ensure there's a final newline:
+  if (cmark_node_is_block(root)) {
+    if (renderer.buffer->size == 0 || renderer.buffer->ptr[renderer.buffer->size - 1] != '\n') {
+      cmark_strbuf_putc(renderer.buffer, '\n');
+    }
   }
 
   result = (char *)cmark_strbuf_detach(renderer.buffer);
--- a/api_test/main.c
+++ b/api_test/main.c
@@ -731,7 +731,7 @@ static void render_commonmark(test_batch_runner *runner) {
   cmark_node *text = cmark_node_new(CMARK_NODE_TEXT);
   cmark_node_set_literal(text, "Hi");
   commonmark = cmark_render_commonmark(text, CMARK_OPT_DEFAULT, 0);
-  STR_EQ(runner, commonmark, "Hi\n", "render single inline node");
+  STR_EQ(runner, commonmark, "Hi", "render single inline node");
   free(commonmark);
 
   cmark_node_free(text);


ioquatix:
While working on this, I wondered if we could upstream some of the cmark-gfm improvements, e.g.

```
#define CMARK_NODE_TYPE_PRESENT (0x8000)
#define CMARK_NODE_TYPE_BLOCK (CMARK_NODE_TYPE_PRESENT | 0x0000)
#define CMARK_NODE_TYPE_INLINE (CMARK_NODE_TYPE_PRESENT | 0x4000)
#define CMARK_NODE_TYPE_MASK (0xc000)
#define CMARK_NODE_VALUE_MASK (0x3fff)

typedef enum {
  /* Error status */
  CMARK_NODE_NONE = 0x0000,

  /* Block */
  CMARK_NODE_DOCUMENT       = CMARK_NODE_TYPE_BLOCK | 0x0001,
  CMARK_NODE_BLOCK_QUOTE    = CMARK_NODE_TYPE_BLOCK | 0x0002,
  CMARK_NODE_LIST           = CMARK_NODE_TYPE_BLOCK | 0x0003,
  CMARK_NODE_ITEM           = CMARK_NODE_TYPE_BLOCK | 0x0004,
  CMARK_NODE_CODE_BLOCK     = CMARK_NODE_TYPE_BLOCK | 0x0005,
  CMARK_NODE_HTML_BLOCK     = CMARK_NODE_TYPE_BLOCK | 0x0006,
  CMARK_NODE_CUSTOM_BLOCK   = CMARK_NODE_TYPE_BLOCK | 0x0007,
  CMARK_NODE_PARAGRAPH      = CMARK_NODE_TYPE_BLOCK | 0x0008,
  CMARK_NODE_HEADING        = CMARK_NODE_TYPE_BLOCK | 0x0009,
  CMARK_NODE_THEMATIC_BREAK = CMARK_NODE_TYPE_BLOCK | 0x000a,
  CMARK_NODE_FOOTNOTE_DEFINITION = CMARK_NODE_TYPE_BLOCK | 0x000b,

  /* Inline */
  CMARK_NODE_TEXT          = CMARK_NODE_TYPE_INLINE | 0x0001,
  CMARK_NODE_SOFTBREAK     = CMARK_NODE_TYPE_INLINE | 0x0002,
  CMARK_NODE_LINEBREAK     = CMARK_NODE_TYPE_INLINE | 0x0003,
  CMARK_NODE_CODE          = CMARK_NODE_TYPE_INLINE | 0x0004,
  CMARK_NODE_HTML_INLINE   = CMARK_NODE_TYPE_INLINE | 0x0005,
  CMARK_NODE_CUSTOM_INLINE = CMARK_NODE_TYPE_INLINE | 0x0006,
  CMARK_NODE_EMPH          = CMARK_NODE_TYPE_INLINE | 0x0007,
  CMARK_NODE_STRONG        = CMARK_NODE_TYPE_INLINE | 0x0008,
  CMARK_NODE_LINK          = CMARK_NODE_TYPE_INLINE | 0x0009,
  CMARK_NODE_IMAGE         = CMARK_NODE_TYPE_INLINE | 0x000a,
  CMARK_NODE_FOOTNOTE_REFERENCE = CMARK_NODE_TYPE_INLINE | 0x000b,
} cmark_node_type;
```

was a nice way to organize the `enum`.

src/node.h
```diff
@@ -86,6 +86,22 @@ struct cmark_node {
 
 CMARK_EXPORT int cmark_node_check(cmark_node *node, FILE *out);
 
+static inline bool CMARK_NODE_TYPE_BLOCK_P(cmark_node_type node_type) {
+	return node_type < CMARK_NODE_TEXT;
```
jgm, line 90: Note that cmark.h defines CMARK_NODE_LAST_BLOCK and CMARK_NODE_FIRST_BLOCK, as well as CMARK_NODE_FIRST_INLINE and CMARK_NODE_LAST_INLINE.  So an easy way to check for blocks is
```c
node->type >= CMARK_NODE_FIRST_BLOCK && node->type <= CMARK_NODE_LAST_BLOCK
```
etc.  This is easy enough that I don't think it's really necessary to create these functions.

jgm:
Looks good. (I made one comment.)

As for the other suggestion, it should probably be considered in a separate issue or PR.
It does seem a good idea.

jgm:
Also, note the linter error in CI.  To keep things simple, let's dispense with defining these extra functions and just use the method noted above in my comment.

ioquatix:
Okay, I will update the PR.

src/node.h
ioquatix: I was trying to establish some level of consistency with the interface introduced by `cmark-gfm`.

jgm: Yes, it may be a good idea in the end.  But I'd prefer to keep that issue separate from this change.

src/render.c
```diff
@@ -148,6 +148,10 @@ void cmark_render_code_point(cmark_renderer *renderer, uint32_t c) {
   renderer->column += 1;
 }
 
+static inline int cmark_is_block(cmark_node *node) {
+  return node->type >= CMARK_NODE_FIRST_BLOCK && node->type <= CMARK_NODE_LAST_BLOCK;
+}
+
```
jgm, line 154: I believe the naming conventions we've been using are that `cmark_*` is used for the functions in the public API.  So I think it may introduce confusion to name this function `cmark_is_block`.  Maybe use `is_block` or `S_is_block` (a conventioned that can be found in e.g. blocks.c.))

ioquatix: Maybe a good reason to extract this into some decent public interface.

jgm: Yes, I'd be in favor of (a) moving these to cmark.h/c, and making them public.
Maybe also (b) [re-organizing the enums the way cmark-gfm does](https://github.com/commonmark/cmark/pull/559#issuecomment-2295678470), to make the tests more efficient.

@nwellnhof can you think of any reason not to do these things?

nwellnhof: > Yes, I'd be in favor of (a) moving these to cmark.h/c, and making them public.

+1

> Maybe also (b) https://github.com/commonmark/cmark/pull/559#issuecomment-2295678470, to make the tests more efficient.

The test is efficient enough. I think this is more about extensibility. Technically, changing the enum values is an ABI break. This shouldn't be a problem as long as we continue to bump the soname with each release and people don't hardcode the constants. But I don't see a good reason to make the change as long as we don't add new node types. If `cmark_is_block` is made public, we should think about deprecating `CMARK_NODE_FIRST_BLOCK` and `CMARK_NODE_LAST_BLOCK`.

nwellnhof: Another function we should consider making public is `S_is_leaf` from `src/iterator.c`.

jgm: OK, let's not change the enums at this point.  But let's add `cmark_is_block`, `cmark_is_inline`, and `cmark_is_leaf` to the public API in cmark.h.

@ioquatix do you want to do this as part of this PR, or should we do that separately and have you rebase?

ioquatix: I think it's better to do it as a separate PR, and I'll rebase this one.

jgm:
Thanks!

PR #588, opened 2025-10-23
Update cmake_minimum_required to 3.14

winterheart:
Recent CMake 4.1 warns that <3.10 support will be deprecated. Update minimum version and simplify some parts of project.

Fix python warnings about regexp escape sequences.

### DIFF
--- a/test/normalize.py
+++ b/test/normalize.py
@@ -18,7 +18,7 @@ import html
 # Normalization code, adapted from
 # https://github.com/karlcow/markdown-testsuite/
 significant_attrs = ["alt", "href", "src", "title"]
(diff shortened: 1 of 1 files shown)


winterheart:
Ubuntu 18.04 reached EOL in 2023, oldest supported Ubuntu release is 22.04 (2027), where cmake 3.22 is available.

nwellnhof:
There's long-term support for Ubuntu, but 18.04 seems old enough. Ubuntu 20.04 on the other hand is still widely used, for example in OSS-Fuzz, even if it's out of standard support. It comes with CMake 3.16, so that's the minimum we should support for a while.

PR #602, opened 2026-03-01
Fix running tests on Python 3.14

### LINKED ISSUE #599: pathological_tests_library .......***Failed
It is not a priority problem but just want to record that the `pathological_tests_library` is failing on fedora.

winterheart:
In Python 3.14 multiprocessing module changed start method from fork to forkserver. Now child processes don't have access to parent's memory objects, and pathological_tests.py fails. This change move required objects (mostly object initialization) into child's area.

Fixes #599


jgm:
thanks!

PR #604, opened 2026-04-17
Avoid stack-overflow and add some defensive conditions

jgm:
These are some changes suggested by Claude in response to a prompt to provide a security audit. They all seem reasonable and correct to me. Some of them aren't strictly necessary, given certain assumptions about how the rest of the code works, but they are probably a good idea in the spirit of defensive coding.

@nwellnhof comments welcome.

### DIFF
--- a/src/blocks.c
+++ b/src/blocks.c
@@ -224,17 +224,17 @@ static void remove_trailing_blank_lines(cmark_strbuf *ln) {
 // Check to see if a node ends with a blank line, descending
 // if needed into lists and sublists.
 static bool S_ends_with_blank_line(cmark_node *node) {
-  if (S_last_line_checked(node)) {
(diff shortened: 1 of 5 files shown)


nwellnhof:
Something like the following seems more readable to me:

```c
static bool S_ends_with_blank_line(cmark_node *node) {
  while (!S_last_line_checked(node)) {
    S_set_last_line_checked(node);
    if (S_type(node) != CMARK_NODE_LIST && S_type(node) != CMARK_NODE_ITEM)
      break;
    if (!node->last_child) 
      break;
    node = node->last_child;
  }
  return S_last_line_blank(node);
}
```

jgm:
Thanks, I agree, and I adopted your suggestion.

PR #606, opened 2026-04-25
Fix python 3 compatibility in normalize.py

### LINKED ISSUE #605: normalize.py is missing a change that the commonmark-spec version has
commonmark/commonmark-spec#815
Fix python compatibility in normalize.py

znewt99:
Add change from the commonmark-spec version of normalize.py
Change from commonmark/commonmark-spec/pull/815
Fixes #605


jgm:
thanks!

PR #613, opened 2026-05-26
Refactor: Use O(1) dereferencing for string emptiness checks instead of strlen()

ismaileneskucuk:
Using `strlen(str) > 0` to check for empty strings incurs an unnecessary O(N) cost, especially inside node rendering functions where it might be called repeatedly. 

Checking the first character `str[0]` is the idiomatic C approach for an O(1) emptiness check.

This PR updates `src/commonmark.c` and `src/main.c` to use this safer and more performant approach, preventing potential performance bottlenecks on large documents with extensive attributes.

### DIFF
--- a/src/commonmark.c
+++ b/src/commonmark.c
@@ -429,7 +429,7 @@ static int S_render_node(cmark_renderer *renderer, cmark_node *node,
         LIT("](");
         OUT(cmark_node_get_url(node), false, URL);
         title = cmark_node_get_title(node);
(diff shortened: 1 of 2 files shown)


jgm:
This makes sense to me, thanks. @nwellnhof  any objections?

nwellnhof:
Looks good to me. Unrelated, but you have to check errno after calling strtol to detect invalid and out-of-range numbers reliably.

PR #615, opened 2026-05-28
Fix: Add errno check to strtol for detecting out-of-range width values

### LINKED ISSUE #614: Check errno after calling strtol
Unrelated, but you have to check errno after calling strtol to detect invalid and out-of-range numbers reliably.

_Originally posted by @nwellnhof in https://github.com/commonmark/cmark/issues/613#issuecomment-4553181190_

ismaileneskucuk:
Resolves #614
Added errno = 0 before parsing and an ERANGE check afterwards to safely exit with an error message when out-of-range width values are provided.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -131,9 +131,13 @@ int main(int argc, char *argv[]) {
           fprintf(stderr, "width value out of range: '%s'\n", argv[i]);
           exit(1);
         }
-        if (unparsed && unparsed[0]) {
+        if (unparsed == argv[i] || (unparsed && unparsed[0])) {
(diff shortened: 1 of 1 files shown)


nwellnhof:
In addition, you have to check that the input string isn't empty, in which case `unparsed[0]` will be 0 despite no conversion being performed. Then we should reject negative or non-positive widths.

ismaileneskucuk:
Updated the logic to use `unparsed == argv[i]` to catch empty inputs.

Also added a check to reject `< 0` values. `0` is kept valid since it is documented as the `nowrap` behavior.

Let me know if anything else is needed.

PR #616, opened 2026-05-28
Refactor: Use snprintf return value instead of redundant strlen

ismaileneskucuk:
`snprintf` naturally returns the number of characters written. Capturing this return value directly avoids the redundant O(N) `strlen` calculation that followed it immediately, providing a zero-cost performance optimization inside the node rendering loop.


nwellnhof:
There could be some theoretical issues on platforms with 64-bit ints where the listmarker buffer isn't large enough. Maybe we should limit the start value to 9 digits in `cmark_node_set_list_start` like it's done in the parser.

PR #617, opened 2026-05-31
Add test changes from commonmark-spec

znewt99:
For `cmark.py`, the last time it was the same was commonmark/cmark@f26d60d and commonmark/commonmark-spec@6aac964.
For `spec_tests.py`, the last time it was the same was commonmark/cmark@54a96ac and commonmark/commonmark-spec@36d2b7e.

This pull request incorporates commits from commonmark-spec. Except, for the change with `decode('utf-8')` since that is what breaks `pathological_tests.py` and `roundtrip_tests.py`.

See also commonmark/commonmark-spec/pull/833


jgm:
thanks

PR #621, opened 2026-06-09
Compile with latest C standard

nwellnhof:
The idea is to catch incompatibilites with newer C standards as early as possible.


nwellnhof:
Enabling C11 or later causes clang-tidy warnings, suggesting to use C11 Annex K functions like `memcpy_s`. These checks should be disabled, but I have no idea where the set of clang-tidy checks is defined in the first place. Running `make lint` locally results in `Error: no checks enabled` which is expected without the `-checks` option or a `.clang-tidy` file. It seems that under Github CI, all `clang-analyzer-security` checks are enabled by default. But I have no idea why.

nwellnhof:
The `clang-analyzer-*` checks were removed from the default checks in clang 22, see

- https://github.com/llvm/llvm-project/issues/146482
- https://releases.llvm.org/22.1.0/tools/clang/tools/extra/docs/ReleaseNotes.html#potentially-breaking-changes

I'll prepare another PR to address this.

PR #627, opened 2026-08-10
Fix reference link title incorrectly kept when followed by trailing garbage

### LINKED ISSUE #468: Link reference definition title that should not be
```
> cmark --version
cmark 0.30.3 - CommonMark converter
(C) 2014-2016 John MacFarlane

> cmark <<EOF
[bla]: /bla
"The title" garbage
[bla]
EOF
<p>&quot;The title&quot; garbage
<a href="/bla" title="The title">bla</a></p>
```

There should be no title attribute on the link here. We have garbage after `"The title"` so it's not one. The [dingus](https://spec.commonmark.org/dingus/?text=%5Bbla%5D%3A%20%2Fbla%0A%22The%20title%22%20a%0A%5Bbla%5D) gets it right.

vjymisal0:
## What

Fixes #468.

`cmark_parse_reference_inline()` in `src/inlines.c` speculatively parses a link title following a reference definition's URL. When the text after the parsed title isn't the end of the line (i.e. there's trailing garbage after the closing quote), the function correctly rewinds and re-checks whether the definition can still be valid by treating the URL's line as ending right after the URL — leaving the "title" text to be parsed as ordinary paragraph content on the next line.

The bug: on that rewind path, the `title` chunk that had already been speculatively parsed was never reset. So even though the code determined the title was invalid (it decided the reference has *no* title, and left `"The title" garbage` as plain text of the next paragraph), it still passed the stale, invalid `title` chunk into `cmark_reference_create()`.

Example:
```
[bla]: /bla
"The title" garbage
[bla]
```

Before this fix:
```html
<p>&quot;The title&quot; garbage
<a href="/bla" title="The title">bla</a></p>
```
The [CommonMark reference dingus](https://spec.commonmark.org/dingus/?text=%5Bbla%5D%3A%20%2Fbla%0A%22The%20title%22%20garbage%0A%5Bbla%5D) (and this same cmark binary via `cmark --version`'s own paragraph output for the discarded title text) shows there should be **no** title attribute here, since `"The title" garbage` is not a valid title specification — only `"The title"` on its own would be.

After this fix:
```html
<p>&quot;The title&quot; garbage
<a href="/bla">bla</a></p>
```

## Fix

Reset `title` to an empty chunk (`cmark_chunk_literal("")`) on the rewind branch, mirroring what the "title didn't match at all" branch already does a few lines below.

## Testing

- Reproduced the bug against current `master` with the CLI (`cmark.exe` built via CMake/MinGW), confirmed the fix resolves it, and confirmed normal (non-garbage-suffixed) titles — both single-line `[bla]: /bla "The title"` and title-on-next-line — still parse correctly.
- Added a regression test to `test/regression.txt` (`Issue #468`) covering this exact case.
- Ran the full CommonMark spec test suite (`ctest -R spectest_executable`) — all pass.
- Ran `ctest -R regressiontest_executable` — all 27 regression cases pass, including the new one.
- `ctest -R html_normalization` and `ctest -R smartpuncttest_executable` also pass.
- `api_test` fails in my local Windows/MinGW build with a DLL-load exception (`0xc0000135`) both **before and after** this change (verified by stashing the fix and rebuilding), so it's a pre-existing local toolchain issue unrelated to this diff, not a regression.


jgm:
Thank you!

PR #628, opened 2026-08-30
Fix thematic break after a link reference definition

### LINKED ISSUE #619: `---` after a link reference definition becomes a paragraph instead of a thematic break
A line of dashes (`---`) immediately following a link reference definition is rendered by cmark as a paragraph (`<p>---</p>`) instead of a thematic break (`<hr />`).

Input:

``` markdown
[f]:r
---
```

Expected:

``` html
<hr />
```

Got:

``` html
<p>---</p>
```

This is specific to `---`. Other thematic breaks—`***`, `___`, and spaced variants `- - -`/`_ _ _`—after a refdef are all correctly rendered as `<hr />`. The bug is also specific to a refdef preceding the dashes; every other non-paragraph block before `---` yields `<hr />`.

BetterAndBetterII:
A line of dashes immediately after a link reference definition was scanned as a setext underline. After the refdef was consumed, that line never fell through to the thematic-break check, so

```markdown
[f]:r
---
```

rendered as `<p>---</p>` instead of `<hr />`. Other thematic breaks (`***`, `___`, `- - -`) already worked because they are not setext underlines.

Only treat a candidate setext underline as a heading when paragraph content remains after resolving reference definitions. Otherwise fall through so `---` can still be a thematic break.

Fixes #619


jgm:
thanks!

Issue #580, opened 2025-07-19
The cmake scripts do not rebuild scanners.c from scanners.re


AndersonTorres:
A cursory reading of [install instructions](https://github.com/commonmark/cmark?tab=readme-ov-file#installing) suggests that both cmake and makefile methods of installation take care of updating scanners.c from scanners.re.

However, the instructions for doing it are not present on cmake scripts, only in the handwritten Makefile.

nwellnhof:
scanners.c and some other pieces of generated code are under version control. Users never have to rebuild these files.

AndersonTorres:
Arguably it doesn't matter for end users, but it is important for developers and package maintainers, especially source-based ones..

Anyway, I will try something after refreshing my memory on cmake.

robertocarlosllamashernandez5-code:
Oye quieres ver otro salgo todo desnudo y con la verga bien parada amor 🙈🔥😋

- ![Image](https://github.com/user-attachments/assets/a3653726-dbad-4590-a02d-2019d0da055a)

- [ ] 

> ``[https://github.com/robertocarlosllamashernandez5-code/Meta-Spatial-SDK-Samples/tree/10f9c7e6bf379d8e251b00432c26dae32ed1a8f1/.github/workflows](url)

nwellnhof:
I don't think CMake offers a clean way to update generated code in the source tree. I'd simply stick with our current Makefile rules.

nwellnhof:
I still don't understand what this issue is trying to solve and I'd suggest to close it.

gemmaro:
For a future reference, [this change for the cmark package definition of GNU Guix](https://codeberg.org/guix/guix/pulls/11169) shows how to build in a bootstrappable way.

---

I think the following two would be helpful for packaging in particular:

1. a way to quickly remove generated files.
   * In the change, it is the `snippet` section.
3. a way to automatically regenerate files when they are missing/removed, due to the first point.
   * In the change, it is the `(add-before 'configure 'generate` part.

I tried to solve the second point and made #630.
I understand this can be bypassed with some adjustments in package recipes, so this is for those who is in a similar situation :)

PR #630, opened 2026-09-11
Add CMake custom commands for generated contents

gemmaro:
Hello, this addresses #580 by using CMake features to generate some support files.

The existing workflow doesn't change.   
`Makefile` tasks for `src/{scanners.c,case_fold.inc,entities.inc}` are kept as is, and the added CMake custom commands are only invoked when `python3`, `re2c`, and `clang-format` are installed.

I added the `.clang-format` file to let clang-format share the consistent behavior across `Makefile` and tool scripts.

I referred the following resources:

* CMake documentation: [`add_custom_command`](https://cmake.org/cmake/help/latest/command/add_custom_command.html) (and its [example usage](https://cmake.org/cmake/help/latest/command/add_custom_command.html#example-generating-files-for-multiple-targets)), [`if` conditionals](https://cmake.org/cmake/help/latest/command/if.html#constant), [`find_program`](https://cmake.org/cmake/help/latest/command/find_program.html), ([`add_custom_target`](https://cmake.org/cmake/help/latest/command/add_custom_target.html) though I didn't use this time)
* Similar file generation with CMake in other projects: [cmigemo](https://github.com/koron/cmigemo/blob/main/dict/CMakeLists.txt)
* [Clang-Format Style Options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)

Thank you,


nwellnhof:
AFAICT, this will rebuild the generated code on each build (if the tools could be found.) This is *not* what we want. Also see my comments in #580.

AndersonTorres:
@gemmaro have you tried to use `add_custom_target` instead?
For us at Nixpkgs and Guix, it would be sufficient. IIRC they generate targets in the backend that can be called independently later via Make or Ninja.

gemmaro:
@AndersonTorres Yes I have.  I think it is a good idea if it is more or less valuable for cmark maintainers as well.

